PR #46, opened 2024-04-23
Debounces inline completion until user stopped typing

AnonymousWalker:
Uses a timer to prevent flooding the text-completion processor. This results in faster response from the model as there are fewer unused queries.

Feel free to make changes as I was using Windows to run it.


ryderwishart (approved):
Really like this approach. Thanks Tony!

PR #47, opened 2024-04-24
Minimal UI settings for editor workspace

AnonymousWalker:
Toggles away some of the menus and panels to avoid overwhelming the user/translator

### DIFF
--- a/src/activationHelpers/contextUnaware/projectInitializers.ts
+++ b/src/activationHelpers/contextUnaware/projectInitializers.ts
@@ -361,6 +361,14 @@ export async function checkForMissingFiles(){
 export async function handleConfig(){
     const config = vscode.workspace.getConfiguration();
 
+    // simplify editor UI when lauching to avoid overwhelming the user
+    config.update('window.menuBarVisibility', 'toggle', vscode.ConfigurationTarget.Global);
+    config.update('window.commandCenter', false, vscode.ConfigurationTarget.Workspace);
+    config.update('workbench.layoutControl.enabled', false, vscode.ConfigurationTarget.Workspace);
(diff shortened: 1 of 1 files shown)


src/activationHelpers/contextUnaware/projectInitializers.ts
```diff
@@ -361,6 +361,14 @@ export async function checkForMissingFiles(){
 export async function handleConfig(){
     const config = vscode.workspace.getConfiguration();
 
+    // simplify editor UI when lauching to avoid overwhelming the user
+    config.update('window.menuBarVisibility', 'toggle', vscode.ConfigurationTarget.Global);
+    config.update('window.commandCenter', false, vscode.ConfigurationTarget.Workspace);
```
ryderwishart, line 366: Okay this looks helpful. I suppose we probably want a function to toggle, and then include some logic for switching between false/true or whatever string is necessary for a given setting.

AnonymousWalker: I believe it will be accessible via "Alt" key press.

```
  // Control the visibility of the menu bar. A setting of 'toggle' means that the menu bar is hidden and a single press of the Alt key will show it. A setting of 'compact' will move the menu into the side bar.
  //  - classic: Menu is displayed at the top of the window and only hidden in full screen mode.
  //  - visible: Menu is always visible at the top of the window even in full screen mode.
  //  - toggle: Menu is hidden but can be displayed at the top of the window via the Alt key.
  //  - hidden: Menu is always hidden.
  //  - compact: Menu is displayed as a compact button in the side bar. This value is ignored when `window.titleBarStyle` is `native`.
  "window.menuBarVisibility": "classic",
```
https://code.visualstudio.com/docs/getstarted/settings#_default-settings

PR #48, opened 2024-05-02
change from draft folder to file folder

BenjaminScholtens:
I've created a project management extension that will generate the folder structure under a files folder instead of the draft folder


dadukhankevin (approved):
It all looks good to me, I don't think there was anything in the Python that was explicitly used /drafts anyway, so it should be good.

PR #71, opened 2025-05-16
Web compatibility updates

sampjvv:
both the webviews build and web build compile successfuly. Still should be tested. Thanks for your patience!


sampjvv:
[Cleaned up web-compatible codex editor. Does not run mocha tests, as the mocha module is not web-compatible. Works as usual when run on vscodium.

PR #191, opened 2025-10-20
Revert Line Numbers from getChapterBasedVerseNumber in CellList

LeviXIII:
Revert Line Numbers from getChapterBasedVerseNumber in CellList as it was breaking subtitle uploads.


BenjaminScholtens (approved):
good enough

Issue #164, opened 2025-09-25
Sliders in Audio upload at not working


BenjaminScholtens:
In that audio uploader the silence threshold slider and some other threshold slider are non functional.

PR #206, opened 2025-10-22
Ben/fix time stamps and audio segments

BenjaminScholtens:
#164 
#205


webviews/codex-webviews/src/NewSourceUploader/importers/audio/AudioWaveform.tsx
```diff
@@ -326,18 +341,34 @@ export const AudioWaveform: React.FC<AudioWaveformProps> = ({
...
+    useEffect(() => {
+        return () => {
+            try {
+                if (playbackSource) playbackSource.stop();
+            } catch (e) {
+                // ignore
+            }
+        };
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [file]);
```
LeviXIII: Should playbackSource be in the dependency array as well? The comment suggests it should, but just want to make sure.

PR #227, opened 2025-10-23
Martin/obs tms roundtrip

Fikitti:
Implemented TMS and OBS roundtrip.

Should be functional, is tested to some extent that files don't lose data during roundtrip.


BenjaminScholtens (changes requested):
Looks like you somehow reverted a change that I made last week.

BenjaminScholtens (approved):

PR #275, opened 2025-10-24
Removed post-sync media cleanup logic to streamline synchronization p…

TimRl:
…rocess.


TimRl (commented):
Simply removed a redundant cleanup

PR #322, opened 2025-11-05
Further fixes for corpus marker renaming

LeviXIII:
* Make sure corpusMarker is updated in .codex and .source files.
* Rename abbreviated Bible names to full names on first load if fileDisplayName is not in file-level metadata.


BenjaminScholtens (approved):
looks good

PR #334, opened 2025-11-10
329 publishing is not working on main

LeviXIII:
- [x] users would click publish
- [x] It should auto select the first group they are a part of and cache the choice 
- [x] when you publish it should publish all the files in the project currently and use lfs to store audio
- [x] We should remove the option to publish privately 

Also adds progress notification when creating project.


BenjaminScholtens (approved):
works good

PR #382, opened 2025-11-21
375 bug merging source cells with audio does not merge the audio

### LINKED ISSUE #374: test needed for merge buttons
We need a test that tests that merge buttons show up in the source when toggle source editing mode is turned on.

### LINKED ISSUE #375: Bug: Merging source cells with audio does not merge the audio
When we toggle source editing mode, it gives us a little button that allows to merge the content of one cell into another in the source. This also duplicates the process on the target side. This all works very well for text, however we did not account for audio. Currently, the audio will not get merged together. We need to update this so that the audio actually gets re-merged using FFmpeg and creates a new audio history in the target merge cell. This means that if you're merging cell 2 into cell 1, cell 2 should be hidden, and cell 1 should end up with a new audio file that's selected in the history. And that audio file contains the audio from both cell 1 and cell 2, but it's actually merged together using FFmpeg.

LeviXIII:
Closes #374
Closes #375  

Fix cell audio merge to correctly include all previously merged audio

- Add audioMerger.ts utility with FFmpeg-based audio concatenation
- Extend mergeCellWithPrevious to merge audio attachments
- Preserve audio properties (startTime, endTime, transcription)
- Add comprehensive test suite
- Handle FFmpeg unavailability gracefully

- When merging cells, read the previous cell's audio attachment directly
from cell metadata instead of searching the filesystem. This ensures
we get the latest merged audio file that includes all previously
merged cells.

- Also adds temporary path handling to prevent file corruption when
output path matches input path, and removes excess debug logging.

- Make tests for merge buttons showing up when toggle source editing mode is on.

- Remove duplicated tests

### DIFF
--- a/src/providers/codexCellEditorProvider/codexCellEditorMessagehandling.ts
+++ b/src/providers/codexCellEditorProvider/codexCellEditorMessagehandling.ts
@@ -1060,6 +1060,10 @@ const messageHandlers: Record<string, (ctx: MessageHandlerContext) => Promise<vo
         await vscode.commands.executeCommand(typedEvent.content.command, ...typedEvent.content.args);
     },
 
+    openLoginFlow: async () => {
+        await vscode.commands.executeCommand("codex-project-manager.openStartupFlow");
+    },
+
     togglePinPrompt: async ({ event }) => {
         const typedEvent = event as Extract<EditorPostMessages, { command: "togglePinPrompt"; }>;
         console.log("togglePinPrompt message received", { event });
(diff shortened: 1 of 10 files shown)


src/test/suite/codexCellEditorProvider.test.ts
```diff
@@ -3520,5 +3991,153 @@ suite("CodexCellEditorProvider Test Suite", () => {
             // Note: Network call to analytics is mocked/optional and tested separately
             assert.ok(true, "Variant selection recorded successfully");
         });
+
+        test("merge buttons show up in source when toggle source editing mode is turned on", async function () {
```
BenjaminScholtens: we need a matching test in the webview tests that checks the the buttons are only visible when the source editing mode it turned on.

PR #390, opened 2025-11-25
378-users-need-to-be-forced-to-update-the-codex-base-app

LeviXIII:
Show version update modal in source webview if syncing and version is below 1.93.

Screenshot 2025-11-25 at 10 27 57 AM


BenjaminScholtens (approved):
looks good works good

PR #401, opened 2025-11-26
Martin/biblica experimental

Fikitti:
Implemented codex cell breaking parser into smaller paragraphs according to <br/> in the Biblica IDML files.

The bible text should not be exported back to the original file anymore.

:D


BenjaminScholtens (approved):
looks good

PR #416, opened 2025-11-28
402 projects need to be saved a unique id in the name of the project on gitlab

LeviXIII:
- ProjectId is now created when naming project.
- ProjectId is added to global state when creating a new project.
- PublishProject webview component stores projectId and appends it to the project name (format: name-projectId) before creating the project.
- Make edited project name appear on Project Setup display in Startup Flow instead of cleaned project name.
- Fix error message that says "Project name cannot be empty" when first opening Create New Project modal.
- Fix projectId mismatch by extracting projectId from folder name when initializing projects in MainMenu (matches StartupFlow behavior)
- Prevent automatic metadata.json creation when Main Menu opens by adding safeguards in updateMetadataFile() and configuration change listener
- Make MainMenu "Initialize Project" button behavior consistent with StartupFlow (check for existing metadata before creating)
- Add loading state to MainMenu Initialize Project button to prevent multiple clicks during initialization
- Make test suite for NameProjectModal.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -468,16 +468,18 @@ export async function activate(context: vscode.ExtensionContext) {
             const pendingCreate = context.globalState.get("pendingProjectCreate");
             if (pendingCreate) {
                 const pendingName = context.globalState.get<string>("pendingProjectCreateName");
-                console.log("[Extension] Resuming project creation for:", pendingName);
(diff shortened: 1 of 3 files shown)


BenjaminScholtens (approved):
words good

PR #495, opened 2025-12-27
492-most-recently-recorded-audio-doesnt-playback-after-syncing

LeviXIII:
- Implement audio attachment availability checks and cache clearing on updates.
- Refactor audio availability logic in CodexCellEditorProvider and update AudioPlayButton and CellContentDisplay components to clear cached audio data when attachments change.


BenjaminScholtens (approved):
works good

PR #538, opened 2026-01-22
System message editing

Luke-Bilhorn:
Ticket 407- Added a system message editing step to the startup sequence, and fixed related bugs.

### DIFF
--- a/src/globalProvider.ts
+++ b/src/globalProvider.ts
@@ -28,6 +28,7 @@ export abstract class BaseWebviewProvider implements vscode.WebviewViewProvider 
 
     // Common webview resolution
     public resolveWebviewView(webviewView: vscode.WebviewView) {
+        console.log(`[WebviewPerf] resolveWebviewView called for: ${this.getWebviewId()}`);
         this._view = webviewView;
 
         webviewView.webview.options = {
--- a/src/providers/registerProviders.ts
+++ b/src/providers/registerProviders.ts
@@ -62,19 +62,19 @@ export function registerProviders(context: vscode.ExtensionContext) {
     const parallelProvider = new ParallelProvider(context);
 
     disposables.push(
-        vscode.window.registerWebviewViewProvider("codex-editor.navigation", navigationProvider),
+        vscode.window.registerWebviewViewProvider("codex-editor.navigation", navigationProvider, { webviewOptions: { retainContextWhenHidden: true } }),
         GlobalProvider.getInstance().registerProvider("codex-editor.navigation", navigationProvider as any),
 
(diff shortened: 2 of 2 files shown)


webviews/codex-webviews/src/components/AdaptiveProjectTitle.tsx
```diff
@@ -19,7 +19,7 @@ export const AdaptiveProjectTitle: React.FC<AdaptiveProjectTitleProps> = ({
     className = "",
     style = {},
     minFontSize,
-    maxFontSize = 16,
+    maxFontSize = 18,
```
LeviXIII: Can we keep the max at 16? 18 is a bit too big.

webviews/codex-webviews/src/StartupFlow/components/SystemMessageStep.tsx
```diff
@@ -0,0 +1,287 @@
...
+                                    display: "inline-flex",
+                                    alignItems: "center",
+                                    justifyContent: "center",
+                                    width: `${ICON_SIZE}px`,
+                                    height: `${ICON_SIZE}px`,
+                                    marginLeft: ICON_LEFT_MARGIN,
+                                }}></i>
+                                <span style={{ paddingLeft: TEXT_LEFT_PADDING }}>Generate Automatically</span>
+                            </span>
+                        )}
```
LeviXIII: It will be more beneficial to have a generated message already waiting for the user. On loading the screen, just have the message generate as well. If the user doesn't like it, they can delete it, but it should be there for them so they have at least something filled in to help with the translations. As much as possible, we don't want this field to be empty unless the user has decided that for themselves.

webviews/codex-webviews/src/MainMenu/index.tsx
```diff
@@ -497,7 +497,7 @@ function MainMenu() {
                                                 }
                                                 className="flex-1 min-w-0"
                                                 minFontSize={9}
-                                                maxFontSize={16}
+                                                maxFontSize={18}
```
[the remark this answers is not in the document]
Luke-Bilhorn: *thumbs up emoji*

types/index.d.ts
```diff
@@ -362,8 +364,10 @@ export type MessagesFromStartupFlowProvider =
     | {
         command: "metadata.checkResponse";
         data: {
-            exists: boolean;
-            hasCriticalData: boolean;
```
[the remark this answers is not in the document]
Luke-Bilhorn: Reinserted

webviews/codex-webviews/src/components/AdaptiveProjectTitle.tsx
Luke-Bilhorn: Set to 16

PR #539, opened 2026-01-22
Finished UI fixes for ticket 511

Luke-Bilhorn:
Addresses ticket 511. Syncing spinner and project titles in settings now display correctly.


LeviXIII (approved):
Looks good. I just changed the max font to 16 instead of 18.

PR #541, opened 2026-01-23
Improved the A/B testing logic for when using a local API and made it…

dadukhankevin:
Codex will not require being logged in when testing the local API because logging in on the frontier server, when it's running locally doesn't work. So now locally it won't require being logged in, the API will return mock results, A/B testing will send to the local API, and when in production, A/B testing will be sent to Frontier API (instead of codex0)


BenjaminScholtens (approved):
Looks good and seems to work well. Just need sign off on the commit I pushed

PR #552, opened 2026-01-29
551 allow for milestone to change on each milestone edit

LeviXIII:
- Milestone now immediately reflects changes when editing instead of waiting for the accordion to close.


BenjaminScholtens (approved):
looks good. works good

PR #577, opened 2026-02-03
Daniel/future of parallel passages

dadukhankevin:


LeviXIII (commented):
The "Retain My Validation" checkbox doesn't seem to work. It checks on and off, but it doesn't actually retain the validation after changing it.


Not sure what "Complete Only" means. Fully validated? Also, not sure if it is filtering correctly.

Image


Clearing the search bar should probably clear out the search results.

Image

Clicking on the Pinned text or number should show me the pinned results. There's no way to view them separately.

Image

I'm getting an error that says I can't navigate away from the current page after running some operations in the Find and replace sidebar. Not sure which actions caused it, but there's no way to navigate even though I don't have a cell editor open. I did find and replace a couple times, so maybe that triggered some bad state?
Screenshot 2026-02-10 at 2 12 19 PM

dadukhankevin:
@LeviXIII 
Retain validation probably broke, thanks!
Complete only just refers to "only show me translation pairs that have both source and target text".
I'll look into the rest of the issues!

PR #578, opened 2026-02-04
Made startup almost instant.

### LINKED ISSUE #579: FastStartup
Startup time was very slow so I moved everything to the background and removed the whole startup screen entirely.

dadukhankevin:
Every startup item that could be put in the background was.


BenjaminScholtens (changes requested):
Image

It's a little two fast for the auth extension to be recognized. We need some sort of loading state that shows we are still waiting to confirm that the is there or not there.

BenjaminScholtens:
Start up is fast enough right now. Users expect to see the loading screen at this point.

PR #600, opened 2026-02-06
Ben/linenumber migration

BenjaminScholtens:
add line number migrations and imporve ui in migration tool


webviews/codex-webviews/src/CodexMigrationToolView/CodexMigrationToolView.css
LeviXIII: Can we just delete this file since styling is handled using tailwind?

PR #604, opened 2026-02-06
587 Line count issue

LeviXIII:
- When using the fallback for cells not in fullDocumentTranslationsUnits (which can happen from a race condition) work with the currently visible cell list.
- Offset the numbers for the list so we get the correct line numbers per section.


BenjaminScholtens (approved):
looks good works good

PR #608, opened 2026-02-09
593 comments UI restructuring

Luke-Bilhorn:
Makes the Comments UI look like Ben's diagram, with comments grouped by file and by milestone, when the sort by "Location in Project" option is enabled in the Comments menu (on now by default, but strict chronological ordering is also still available). 

This should make PRs for 540 and 549 unneeded, methinks, because it incorporates all of their changes as well.


BenjaminScholtens (approved):
looks good

PR #639, opened 2026-02-18
636 using the sparkle button inside the textcelleditor immediately closes the editor after the llm generation

LeviXIII:
- Fix type on setContentToScrollTo function for search navigate to cell.
- Keep editor open after sparkle/LLM refresh; clear dirty state only when leaving page
  - Use UnsavedChangesContext at top of component with other context
  - On page load/cell-page: only clear contentBeingUpdated and unsaved state when the currently edited cell is not on the loaded page
  - When the edited cell is on the page (e.g. after sparkle LLM refresh), preserve the editor and sync only cellLabel and cellTimestamps from the page; do not overwrite cellContent so Quill keeps authoritative content and save does not persist empty/stale data
  - Call setUnsavedChanges(false) when clearing editor state so navigation is not blocked
- Hide Save/Discard when opening editor with existing content
  - Editor: ignore text-change for dirty/unsaved until init is done (100ms after setSelection/focus).
  - TextCellEditor: only set unsaved when user has edited (ref + gate combine-dirty effect); reset ref on cell change.
- Adjust timing on tests to account for change in timing for initial values of text editor.
- Remove unnecessary test.

### DIFF
--- a/docs/AB_TESTING.md
+++ b/docs/AB_TESTING.md
@@ -3,7 +3,7 @@
 A/B testing in Codex shows two translation suggestions side‑by‑side once in a while so you can pick the better one. This helps us learn which retrieval and prompting strategies work best without slowing you down.
 
 ## How it works
-- Triggering: Tests run at random with a small probability (default 15%).
+- Triggering: Tests run at random with a small probability (default 1%).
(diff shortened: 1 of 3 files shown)


BenjaminScholtens (approved):
looks good works good

PR #738, opened 2026-03-11
684 search and replace v2 fixes and improvements

Luke-Bilhorn:
Added these 3 fixes that Matt requested!

I'll get the one Ben asked for in a little bit.


Luke-Bilhorn:
Ok, so the merge actually did not break stuff. 

What happened was that I thought I could safely keep all of the edits from main and my branch in Comments View, and then when I tested it I got build errors. I kinda panicked and reverted it, but then when I worked through the whole thing on a couple other branches I realized that it was just a package issue on my machine. So, I reapplied the merge and tested it, and it all works now so far as I can find.

LeviXIII (commented):
I think you will have to cherry pick the changes that only pertain to the search and replace. This PR undoes the improvements to the comments, so it looks like your merge went bad.

Besides that, the search and replace is working better. A couple of things to change still:

1. When searching, only the target text gets highlighted and not the source

Image

2. When clicking on the result card, I do get highlighted text which is good, but every time I click another cell in the target the result card cell still gets highlighted.

https://github.com/user-attachments/assets/15a98d42-d513-4e84-a88c-ba66c9599026

3. We should change the 1 of 14 of 55 to something more coherent. I'm thinking that there are 55 different milestones with the word, so maybe we can convey that differently. Like 1 to 14  -  Page 1? 

Image

LeviXIII:
https://github.com/genesis-ai-dev/codex-editor/pull/745

Issue #790, opened 2026-03-24
HMethiopia porject: USFM files not uploading - Client is experiencing an issue uploading USFM flie


jmavescodex-arch:
Problem: despite using the correct upload method the USFM file is not uploading for the client

<img width="2866" height="1758" alt="Image" src="https://github.com/user-attachments/assets/9bd5cd4e-b417-4e92-9927-f3a1ab4750b6" />

PR #813, opened 2026-03-30
806 fix the problems with importing target files

Fikitti:
Renewed the way we handle the import of target files and mapping especially verse mapping for USFM file to fix a problem for client from ticket #790


LeviXIII (approved):

webviews/codex-webviews/src/NewSourceUploader/NewSourceUploader.tsx
```diff
@@ -287,13 +282,38 @@ const NewSourceUploader: React.FC = () => {
...
 
-                console.log("Received project inventory:", inventory);
+                const initialIntent: ImportIntent | undefined = message.initialIntent;
 
-                setWizardState((prev) => ({
-                    ...prev,
-                    projectInventory: inventory,
-                    isLoadingInventory: false,
-                }));
+                console.log("Received project inventory:", inventory, "initialIntent:", initialIntent);
```
Luke-Bilhorn, line 287: make it a debug?

Luke-Bilhorn (commented):
I don't want to gum the wheels, but should we make these log statements into debug statements?

Luke-Bilhorn (approved):
Go pull out those logs (there are seven) but other than that, looks good!

Exporting preserves changes to files.

webviews/codex-webviews/src/NewSourceUploader/NewSourceUploader.tsx
LeviXIII: Good catch. I think we will make a separate pass to remove all our console.logs that we don't need.

PR #818, opened 2026-03-30
Dev tools sql with lukes changes - 6

TimRl:


src/projectManager/projectExportView.ts
```diff
@@ -734,73 +785,78 @@ function getWebviewContent(
...
                 function exportProject() {
-                    const formatToSend = selectedFormat || (selectedAudio ? 'audio' : null);
+                    let formatToSend = selectedFormat || (selectedAudioMode ? 'audio' : null);
                     if (!formatToSend || !exportPath || selectedFiles.size === 0) return;
                     const options = {};
-                    if (formatToSend === 'usfm') options.skipValidation = document.getElementById('skipValidation')?.checked;
-                    // Audio is now a separate toggle that may be combined with other export formats
-                    if (selectedAudio) options.includeAudio = true;
-                    if (selectedAudio) options.includeTimestamps = document.getElementById('audioIncludeTimestamps')?.checked;
+                    if (formatToSend === 'usfm-no-validate') {
```
LeviXIII, line 838: Right now, I can only see the "Generate USFM Without Validation" option if I click on "Generate USFM". You might as well just group it with the other options in "Text and Markup Export Options" as it should be a valid choice from the start. Just be sure that it gets hidden whenever the "Generate USFM" option is hidden as well.

LeviXIII (approved):

PR #820, opened 2026-03-30
778 if a vtt or srt file is not valid in some way we should give a warning during import

### LINKED ISSUE #778: If a vtt or srt file is not valid in some way we should give a warning during import
Specifically, the chosen has been generating some vtt files using the chatgbt or something like that. And sometimes it messes up the timestamps. So line 3 will have a timestamp that belongs at the end of the file instead of in between a timestamp that belongs in between line 2 and line 4. So that's one thing that we should definitely show an error about or some sort of warning when they are importing a vtt.
We could also show any other parsing errors we can think of or that an LLM can think of, but specifically this non-sequential parsing error has come up a few times now, and we should be giving our users a warning when they're importing.

Fikitti:
Added a warning for user when importing corrupted vtt / srt files with not sequential time stamps which result and corrupted outcomes.


BenjaminScholtens (approved):
looks good works good

BenjaminScholtens (changes requested):
let's just make it give a warning any time things are out of order.

BenjaminScholtens (approved):
looks good works good

Luke-Bilhorn (approved):
Looks good works good

PR #821, opened 2026-03-31
Dev tools sql with lukes changes improved ext test yaml

TimRl:


BenjaminScholtens (approved):
seems to work faster

TimRl:
> seems to work faster

But, I tried to upgrade the node version and that broke things. We need to upgrade to node 24 though

LeviXIII (approved):

PR #829, opened 2026-03-31
[Ext Pinning] Feat: perfer remote pin over local

JonahBraun:
Fixes a bug where codex-editor wouldn't load due to pin mistmatch when we are trying to use a new pin coming from a remote project sync.


JonahBraun:
outdated

PR #836, opened 2026-04-02
833 do not allow the option to export round trip if in stream only mode

### LINKED ISSUE #833: Do not allow the option to export Round Trip if in Stream Only mode
The yellow tag is not enough for Matthew apparently -_-;

Fikitti:
The Round-trip option is greyed out for the stream only and is only available for the other instance options. There is also a yellow warning explaining why it's not available in stream only and what users should do to Round-trip their files.


LeviXIII (approved):

Luke-Bilhorn:
Why are we rewriting our entire package lock?

Luke-Bilhorn (approved):
It works!

PR #839, opened 2026-04-04
Fix project update and swap array deletion at merge

### LINKED ISSUE #775: Project updating deletes the project update if it was not synced locally (something in the merge doesn't happen) and certain changes get discarded
When performing a project update, the array might get cleared, when it should simply get displayed as executed

TimRl:


LeviXIII (approved):
Tested with Tim and works well.

Luke-Bilhorn (approved):

PR #842, opened 2026-04-07
827 Audio Export doesn't work when exporting without timestamps

### LINKED ISSUE #827: Audio export Bug in dev
Image

If we select this audio export option the wav files exported are corrupted. 

There is also an issue with the getting the correct count when naming files on an audio export.

Image

LeviXIII:
Closes #827 

### Summary
- Fix the one off naming issue on exported audio filenames.
- Keep original extension for files instead of forcing them to .wav when exporting.

### To Test
- [ ] Try to export an audio file without timestamps and make sure it opens correctly
- [ ] Make sure the numbering of each file starts at 1 and goes sequentially from there


dadukhankevin (approved):
So, I looked at the code. Looks good. I'm at the conference so I didn't test it out. But I can't imagine too many things could go wrong because it's like not a huge PR. So anyway, I approve.

TimRl (approved):

PR #844, opened 2026-04-09
840 export options showcase

Fikitti:
Added feature for Export Option Preview when selecting importer type. Selecting importer is now 2 step verification, where export options are visible on the side. Green ones represent compatible options with chosen importer, when user also hovers on the options, they get more information about the exporter. User can hit Cancel to close the export option view or Continue to go to file upload for chosen importer as usual.


Luke-Bilhorn (commented):
When the view that the importer flow is opened into is less than about 60% the width of my screen, the export options showcase displays at the very bottom of the scrollview, where users are likely to have trouble finding it. 

Could we place this up within the grid of importers itself? So I click "Audio", and then the exporter showcase options panel opens up beneath the line of cards containing the audio card. 

This is a bit more involved, but the AI can write that up pretty quick. If we want to leave it as is for now and do this on another ticket, I also understand that.

Luke-Bilhorn (commented):
Also, I see that turndown is imported but never used. Can we get rid of it?

Luke-Bilhorn (dismissed):
The showcase now appears on the side, but this is workable. Maybe should be moved into the same column as the importer type cards in the future, but this is hopefully at least a visible solution.

LeviXIII (changes requested):
Same comment as Luke, but do we need Turndown in the package.json?

LeviXIII (approved):

Luke-Bilhorn (approved):
I previously reviewed and approved; Matt said it still works after last set of changes.

PR #855, opened 2026-04-10
Ben/fix sparkle

BenjaminScholtens:
Found an error where llm completions won't trigger properly for some abnormal cases. This adds a proper check.


LeviXIII (dismissed):
Worked for me.

PR #858, opened 2026-04-13
Project settings validation increment fix

### LINKED ISSUE #853: Fix audio and text validator incrementing (still buggy)

TimRl:
### Problem

When clicking the +/- buttons on the Required Validations counter (text or audio), the displayed value would jump erratically — e.g., going from 1 to 6, then toggling between 2 and 6. Clicking at ~0.5s intervals made it worst, as each click's round-trip response collided with the next click. Additionally, rapid clicking produced a separate metadata edit entry for every intermediate value instead of one clean edit.

### Root Cause

1. **Write ordering race:** The handler `handleSetValidationCountDirect` updated config first, then metadata. The config update fired `onDidChangeConfiguration` listeners that called `refreshState()`, which read metadata before it was updated — pushing stale values back to the UI.

2. **Premature optimistic clearing:** The webview cleared its optimistic local state immediately when the server confirmed the correct value. If a stale `stateUpdate` arrived after that (from a racing listener), the display regressed to the old value.

3. **Debounce too short:** The 150ms debounce meant clicks spaced ~0.5s apart each produced a separate round-trip, metadata write, and edit entry — amplifying both the race condition and edit history noise.

### Fix

1. **Provider — reorder writes (mainMenuProvider.ts):** Write metadata first, then config. By the time `onDidChangeConfiguration` listeners fire and read metadata, it already has the correct value.

2. **Webview — settlement delay (MainMenu/index.tsx):** Add a 500ms delay before clearing optimistic local state after server confirmation. This absorbs any late stale `stateUpdate` messages. A functional updater ensures new clicks during the window are preserved.

3. **Webview — increase debounce to 1.2s (MainMenu/index.tsx):** Only the final value is sent to the provider after the user stops clicking, producing a single metadata write and one clean edit entry. The UI stays instant throughout via optimistic local state.

### How to Test

1. Open a project in Codex Editor and navigate to the Main Menu sidebar
2. Find the "Required Validations" section with Text and Audio counters
3. **Slow clicks (~0.5s apart):** Click + repeatedly at roughly half-second intervals. The counter should increment by 1 each click without jumping or reverting
4. **Rapid clicks:** Click + as fast as possible 5-6 times. The counter should end at the correct final value without intermediate flickering
5. **Decrement:** Repeat steps 3-4 with the - button
6. **Edit history:** After going from e.g. 1 to 5 via rapid clicks, check metadata.json — there should be one edit entry (1 → 5), not four intermediate entries
7. **Reload:** After changing values, reload the window and confirm the values persisted correctly
8. **Both counters:** Repeat for both Text and Audio validation counters


LeviXIII (dismissed):
Nice job!

LeviXIII (approved):

PR #860, opened 2026-04-13
Improved Markdown Rountrip

### LINKED ISSUE #846: Improve the Markdown round-trip logic
The current markdown round-trip is not using the original file for rebuilding and is not preserving all the markdown tags and stylings as we would want it to, since it's labeled as "Round-trip" importer.

Fikitti:
Revisited markdown import/export pipeline, and making sure the markdown files are round-triping as intended now with no structure loses. We now also preserve the original file in the "/originals" and rebuild the file accordingly. It is also the "enforceHTMLstructure" importer that pings user when original html strucutre of the cell is missmatched.


LeviXIII (approved):
Looks good. Worked well for me.

PR #864, opened 2026-04-14
721 main we should be saving the users current codex binary version in the metadata

### LINKED ISSUE #721: We should be saving the users current codex binary version in the metadata
We need to know if all the users have the latest binary so we know what project will be compatible with targeted deploy. 

We should start saving the current users codex binary version to the metadata file when they open the project or when they sync.

it could look something like this:

```
...
"users": [
{
userName: "matt_test",
codexVersion: "109.123.456",
updatedAt: 12989389382
},
...
]
...
```

LeviXIII:
Closes #721 

### Summary
- Create a users array with ProjectUserVersionEntry which contains the codexVersion.
- Write tests for changes

### To Test
- [x] Check if the user array populates correctly on reload and sync
- [x] Check user array with another person to check merging strategy when syncing

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -582,6 +582,12 @@ export async function activate(context: vscode.ExtensionContext) {
                 } catch (error) {
                     console.warn("[Extension] Error ensuring extension version requirements:", error);
                 }
+
+                try {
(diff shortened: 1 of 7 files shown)


BenjaminScholtens (approved):
works good

Luke-Bilhorn (approved):
Users array extends correctly!

PR #868, opened 2026-04-15
861 text only paste option

### LINKED ISSUE #861: Text Only Paste Option
We are having issues where people are copying and pasting special text into the TextCellEditor (eg. has coloured background because of styling tags). We need to have an option to paste just the text into the editor so that they won't see any of the styling of what they copied.

LeviXIII:
### Summary
- Adds a Paste as plain text option for the Quill-based cell editor so pastes use clipboard plain text only (no rich HTML / styling). The choice is persisted per workspace (same pattern as the preferred editor tab), so it survives reopening the editor.

### What changed
- Webview ↔ extension: New messages getPasteAsPlainText, setPasteAsPlainText, and pasteAsPlainTextPreference (see types/index.d.ts).
- Persistence: Workspace state key codex-editor-paste-as-plain-text via workspaceStateUtils.ts, wired through codexCellEditorProvider.ts and codexCellEditorMessagehandling.ts.
- Editor: Capture-phase paste handler runs before Quill’s handler when plain-text mode is on (preventDefault + stopPropagation), then insertText with text/plain. A ref keeps the handler in sync when the toggle changes without remounting Quill.
- TextCellEditor: Checkbox + character count on one row (justify-between); character count is lifted from Editor via optional onCharacterCountChange.
- Tests: webviews/codex-webviews/src/CodexCellEditor/__tests___/Editor.pasteAsPlainText.test.tsx (Vitest).

### Testing
- [x] Open a .codex cell in TextCellEditor.
- [x] Character count: Confirm the row under the editor shows “N characters” on the right and updates as you type.
- [x] Rich paste (toggle off): Uncheck Paste as plain text.
- [x] Copy formatted text from a browser or Word (bold, links, etc.).
- [x] Paste into the cell and confirm formatting still appears (Quill default).
- [x] Plain paste (toggle on): Check Paste as plain text.
- [x] Paste the same rich source; content should match plain text only (no doubling, no extra markup).
- [x] Confirm the checkbox is still on and pastes remain plain after enabling the checkbox, close the cell editor / reload the webview or reopen the notebook, reopen a cell.
- [x] Selection replace: With plain mode on, select part of a verse, paste; only the selection should be replaced by plain text (no duplicate full line).

### DIFF
--- a/src/providers/codexCellEditorProvider/codexCellEditorMessagehandling.ts
+++ b/src/providers/codexCellEditorProvider/codexCellEditorMessagehandling.ts
@@ -555,6 +555,27 @@ const messageHandlers: Record<string, (ctx: MessageHandlerContext) => Promise<vo
         }
     },
 
+    getPasteAsPlainText: async ({ webviewPanel, provider }) => {
+        try {
+            const enabled = provider.getPasteAsPlainText();
+            provider.postMessageToWebview(webviewPanel, {
+                type: "pasteAsPlainTextPreference",
+                enabled,
+            });
+        } catch (error) {
+            console.error("Error getting paste-as-plain-text preference:", error);
+        }
+    },
+
+    setPasteAsPlainText: async ({ event, provider }) => {
+        const typedEvent = event as Extract<EditorPostMessages, { command: "setPasteAsPlainText"; }>;
+        try {
+            provider.updatePasteAsPlainText(typedEvent.content.enabled);
+        } catch (error) {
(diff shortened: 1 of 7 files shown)


TimRl (approved):
Good stuff!

TimRl (changes requested):
It's working now, might have just been a bug (one-off)

But only one formatting style can be applied to the pasted content (could be a quill editor issue though)

TimRl (dismissed):

TimRl (approved):

PR #871, opened 2026-04-15
863 ebibles with ranges are not being imported correctly

### LINKED ISSUE #863: eBibles with Ranges are not being imported correctly.
Image

There seems to be an issue with some ebibles that are from Indonesian (ind). I think it's because they have ranges. If you look at the image you will se the source skips verses but the target does not.

Fikitti:
New way of creating target files when importing source, the source is blanked (stripped) of text data and audio data and set as a target to match the exact structure and cell count etc.


LeviXIII (approved):
I tried every type of import to see if they came through ok and they did.

PR #882, opened 2026-04-16
Suppress auto-sync when admin pin intent is active

JonahBraun:
After an admin pins an extension, the Conductor sets `adminPinnedExtensions` in workspace storage and reloads into the new profile. The window reload re-triggers the initial workspace sync (and the 5-minute interval auto-sync), immediately pushing the pin to remote before the admin can sanity-test the pinned version.

## Changes

- `src/projectManager/syncManager.ts` — In `executeSync`, block all non-manual syncs when `codex.conductor.hasAdminPinIntent` returns true. Covers both the initial workspace sync on project open and the 5-minute interval auto-sync. Explicit syncs (`isManualSync = true`) are unaffected.

## Test plan

- [x] Pin an extension via Manage Extension Pins; confirm window reloads into the pinned profile
- [ ] After reload, confirm no sync fires automatically (check logs for "Admin pin intent active")
- [ ] Confirm the 5-minute auto-sync is also suppressed while pin intent is active
- [ ] Run an explicit sync (command palette → Frontier: Sync Changes); confirm it proceeds normally and clears the admin intent
- [ ] After intent is cleared, confirm the next window reload triggers auto-sync as usual


JonahBraun:
Pre-release: 0.24.0-pr882-180d29f9 https://github.com/genesis-ai-dev/codex-editor/releases/tag/0.24.0-pr882-180d29f9

BenjaminScholtens (approved):
works good

PR #894, opened 2026-04-21
782 improve docx importing with recursive text splitting to avoid excessively large cells

### LINKED ISSUE #782: Improve docx importing with recursive text splitting to avoid excessively large cells
When importing a .docx file, block styles (e.g., paragraphs) provide the main cell divisions. Some paragraphs can be very long, however, making validation and drafting unwieldy and unclear, diluting our main value proposition. 

Therefore, we can implement recursive text splitting until a typical target cell length is found.

Basically, split on new paragraphs. If the length is longer than the target cell length, split by newlines, then split by `.|!|?`, then split by minor stops, then split by whitespace (or some variation of this pecking order).

This will ensure we extract usefully sized validation units.

Luke-Bilhorn:
Closes #782 

More we could do here; can update later. 

## Testing Checklist

- [x] Confirm the **"Advanced Settings"** section appears for `.docx` imports, with an **"Ideal cell length"** box pre-filled to **160** and helper text ("Set to 0 to disable splitting").
- [x] Import a long multi-sentence paragraph (400+ chars) → it becomes **multiple sentence-sized cells** instead of one giant cell.
- [x] Import short paragraphs (under ~175 chars) → each stays as **one cell**, nothing split.
- [x] Read the new cells in order → text matches the original exactly, **nothing lost or duplicated**.
- [x] Confirm splitting only affects DOCX (the setting does not appear for other importers).

**Ideal cell length setting**
- [x] Set length to **80** vs **300** on the same file → 80 gives **more/smaller** cells, 300 gives **fewer/larger** cells.
- [x] Set length to **0** → splitting fully disabled, every paragraph stays one cell.
- [x] Box rejects negative/non-numeric input without breaking.

**Split-point quality (pecking order)**
- [x] Normal sentences → breaks land at **sentence ends** (`.` `!` `?`), not mid-sentence.
- [x] Run-on with commas but no periods → breaks land at **commas/semicolons**.
- [x] No punctuation at all → breaks land **between words**, never mid-word.
- [x] Numbers like `3.14159` are **never split** inside the number.
- [x] No awkwardly tiny leftover cells are produced.
- [x] Slightly-over-limit or comma-only paragraphs may stay as one cell — confirm this is **expected, not a bug**.

**What should NOT split**
- [x] **Tables** keep existing behavior (cells not run through the splitter) and still map correctly.
- [x] **Empty paragraphs / empty table cells** are preserved, not dropped.

**Formatting**
- [x] Bold/italic/colored text spanning a split → formatting **preserved** in each resulting cell.
- [x] Paragraph alignment/indent styling retained after splitting.

**Round-trip export**
- [x] Translate several split cells, then export → split pieces **rejoin into the original single paragraph, in correct order**.
- [x] Table-cell content still exports to the right place.
- [x] Export with some split cells left untranslated → result is coherent, no crash.

**Language / locale awareness**
- [x] English source: abbreviations like "Mr. Smith" don't break awkwardly when avoidable.
- [x] No-space scripts (Thai/Japanese/Chinese) source: long paragraphs still split into reasonable chunks.
- [x] Unusual/unsupported language tag → import still works (quiet fallback, no error).

**Downstream usability**
- [x] Each split cell can be **drafted** independently (AI/manual).
- [x] Each split cell can be **validated** independently.

**Stability**
- [x] Close and reopen the imported notebook → cell structure consistent.
- [x] Large `.docx` (many long paragraphs) imports without freezing or performance regression.

**Regression**
- [x] No console errors during import; progress reaches 100% and the notebook opens.
- [x] A normal short-paragraph document imports exactly as before.


LeviXIII (approved):
UI options are great (Advanced options) and work well.

Just a note on tables: they will create empty cells in the source. This is the same behaviour as before. When exporting on a roundtrip, everything works as it should and the tables are still in place (without translation as they are ignored). Things are working as they did originally and the cell splits don't reflect in the exported file.

PR #900, opened 2026-04-22
823 warn users about exports that will be empty

### LINKED ISSUE #823: Warn users about exports that will be empty
Currently when exporting, we export both text and audio folders that have nothing in them. Text files might be in there, but they'll be empty, and folders might be empty if it's just audio. We should probably warn the user and possibly not include the folder.

Fikitti:
New exporter UI features with file labels ("No content", "Text only", "Audio only", "Text + Audio").
The user can not check "No content" files and also cannot select "Text only" and "Audio only" files at the same time.

Select all files checkbox at the top of section was moved and changed to two checkboxes, saying "All text", "All audio".

When user will try to export a file without audio translations, and select also audio export (when they have Text only and Text + Audio file selected, the audio option is available), they will be prompted with warning popup saying which files will have empty folder output for the selected exporter.

Also after the export, the empty folder will contain a file "NOTICE.txt" saying why this folder is empty.

How to test:
- [x] have multiple files in different states of translations (some with no content, some with only text, some with only audio and with both)
- [x] try selecting files with the checkboxes (some options will grey out)
- [x] try exporting "Text only" and "Text + Audio" with also Audio export option and the same for "Audio only" with text option export, see if the warnings are working and if the NOTICE file was created

### DIFF
--- a/src/exportHandler/exportHandler.ts
+++ b/src/exportHandler/exportHandler.ts
@@ -34,6 +34,7 @@ import { exportCodexContentAsPlaintext } from "./plaintextExporter";
 import { exportCodexContentAsXliff } from "./xliffExporter";
 import { exportCodexContentAsUsfm } from "./usfmExporter";
 import { exportCodexContentAsHtml } from "./htmlExporter";
+import { analyzeNotebookContent } from "../projectManager/utils/exportViewUtils";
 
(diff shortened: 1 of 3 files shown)


BenjaminScholtens (approved):
looks good works good

PR #901, opened 2026-04-22
Audio count down feature

LeviXIII:
PROJECT TEST FOR CODEX BETA: pr-audio-count-down-feature

**NOTE: Some commits were cherry picked from the audio dubbing branch.**

### Summary

This work improves **audio recording** in the cell editor by adding a **short countdown before recording starts**, clearer **button states** (mic vs countdown number vs stop), and a **visual bar** that compares **how long you recorded** to the **length implied by the cell’s video timestamps**.

Recording only starts when the cell has **start and end timestamps**; otherwise the record action stays disabled and a simple message explains why.

### What changed

- **3–2–1 countdown** before the microphone turns on, so you are not caught off guard when recording starts.
- **Cancel during countdown**: the same control used to stop recording can cancel the countdown before recording begins.
- **Progress bar while recording** (Audio Recording flow): fills from green toward yellow/red as you approach (and pass) the timestamp-based target length, with labels for elapsed time vs “Timestamp Length.”
- **After you have audio** (waveform view): a similar bar compares **decoded audio duration** to the same timestamp target when that target exists.
- **Refactor of save logic**: saving the recorded blob to the cell is wrapped in `useCallback` and called via a ref from the recorder’s `onstop`, so behavior stays correct as dependencies change.
- **Timers cleaned up** on unmount and when opening the recorder again, to avoid stray intervals.
- **Tests**: the locked-cell recording test now finds the button by its **accessible name** (title), matching the UI.
- **`package-lock.json`**: root `version` aligned with `package.json` (0.25.0).

---

## Test checklist

Use a cell **with** video start/end timestamps and one **without**.

### Recording and countdown

- [x] Open the **Audio Recording** UI for a cell that **has timestamps**. The large record button is enabled and shows the **mic** icon.
- [x] Click **Start Recording**. You see **3**, then **2**, then **1** on the button; the button is **green** during countdown; tooltip mentions starting soon.
- [x] After **0**, recording actually begins: button turns **red** and pulses, icon is **stop (square)**.
- [x] During countdown, click again (or use the same stop control): countdown **cancels**, no recording starts, **no** microphone prompt if you cancelled before `getUserMedia`.
- [x] Let countdown finish, then **stop** recording: file saves as before, waveform appears, no obvious errors in the webview.

### Progress bar (recording tab)

- [x] While recording, the bar under the button shows **elapsed** vs **timestamp length**; color shifts toward yellow/red as you get near or past the target (spot-check visually).

### Waveform tab (after audio exists)

- [x] With audio on a timestamped cell, the comparison bar appears under the waveform (actual duration vs target) when `targetDuration` applies; no crash if decoding fails (may show “—” for actual).

### Regression / stability

- [ ] **Upload audio** and **transcription** controls still work as before.
- [ ] **Re-open** recorder from waveform (“record again” path): no duplicate countdowns or stuck timers.
- [ ] Navigate away or close the editor while countdown or recording is active: no runaway timers (best-effort smoke test).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "codex-editor-extension",
-    "version": "0.22.0",
+    "version": "0.25.0",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "codex-editor-extension",
-            "version": "0.22.0",
+            "version": "0.25.0",
             "license": "MIT",
             "dependencies": {
                 "@types/csv-parse": "^1.2.5",
--- a/webviews/codex-webviews/src/CodexCellEditor/AudioWaveformWithTranscription.tsx
+++ b/webviews/codex-webviews/src/CodexCellEditor/AudioWaveformWithTranscription.tsx
@@ -41,6 +41,7 @@ interface AudioWaveformWithTranscriptionProps {
     onShowRecorder?: () => void;
     audioValidationPopoverProps: AudioValidationPopoverProps;
     validationStatusProps: ValidationStatusIconProps;
+    targetDuration?: number | null; // Target duration (in seconds) derived from cell timestamps.
 }
 
 const AudioWaveformWithTranscription: React.FC<AudioWaveformWithTranscriptionProps> = ({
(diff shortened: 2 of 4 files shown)


BenjaminScholtens (approved):
looks good. Works good

PR #905, opened 2026-04-23
Binaries in versioned folders and fixing download options, retries and more

TimRl:
# Binary Management Overhaul

## Summary

Brings SQLite and FFmpeg binary storage in line with the already-versioned Git binary, introduces a consistent per-session retry system across all three tools, and overhauls the "Tools Status" webview UI. The cross-session `globalState` retry counters that permanently blocked re-download attempts have been removed.

---

### Changes

**Versioned storage & migration**
- `sqliteNativeBinaryManager.ts`: Binaries now stored at `{globalStorage}/sqlite3-native/{SQLITE3_VERSION}/node_sqlite3.node`. On first run, any existing flat binary is migrated into the versioned subfolder rather than re-downloaded.
- `ffmpegManager.ts`: Same pattern — `{globalStorage}/ffmpeg/{platform-version}/ffmpeg`. Migration reads the version from `PLATFORM_MAP`, not from `ffmpeg -version` output (which can differ from the package version).

**Consistent retry logic**
- All three tools now run up to `MAX_FULL_RETRIES = 2` full download attempts per session with exponential backoff and detailed progress notifications.
- Removed the `globalState`-backed `binaryRetryCount.*` counter system entirely from `binaryIntegrityUtils.ts`. Tools that previously exhausted their persistent counter and silently stopped retrying will now attempt a fresh download on every extension activation.

**Startup behavior**
- `extension.ts`: The "Tools Status" webview no longer auto-opens on startup. A `showErrorMessage` toast is shown only for critical SQLite failures.
- Tool status logging now reports one of five distinct states: `native`, `available but using fallback`, `missing and using fallback`, `available but forced to fallback`, `missing and forced to fallback`.

**Tools Status UI (`MissingToolsWarning.tsx` / `MissingToolsWarningProvider.ts`)**
- Download button uses a state machine: `idle → downloading → (Attempt N / failed) → idle`.
- A tool in `builtin` mode with a missing native binary now shows `error` severity and a visible "Download and Install" button (previously it was hidden).
- On a successful download, the tool mode auto-switches from `builtin` → `auto` so the newly installed binary is used immediately without a manual settings change.

**Types**
- `types/index.d.ts` and `StartupFlow/types.ts` updated for new tool state and download result message shapes.

---

### Changes

**New `gitBinaryManager.ts`**
- Downloads and manages the `dugite-native` Git binary at `{globalStorage}/git/{DUGITE_NATIVE_TAG}/`.
- Verifies integrity via SHA-256 marker file on every startup.
- `MAX_FULL_RETRIES = 2` outer retry loop with exponential backoff, plus separate `MAX_API_ATTEMPTS = 2` and `MAX_DOWNLOAD_ATTEMPTS = 2` for the GitHub API and tarball fetch.
- All cross-session retry counter logic removed; `deleteGitBinary` and `retryGitBinaryDownload` no longer call any reset functions.

**Native + fallback Git routing**
- `dugiteGitNative.ts`: Full native git implementation via `dugite`.
- `isomorphicGitAdapter.ts`: Isomorphic-git fallback for when the native binary is unavailable.
- `dugiteGit.ts`: Router that selects native vs. fallback at runtime.

**Offline handling**
- Connectivity check changed from `api.github.com` (which returns non-`ok` under rate limiting, treating it as offline) to `github.com`, with timeout increased to 5 s.

---

### Binary versioning & migration
- [ ] Fresh install: SQLite, FFmpeg, and Git binaries download into their versioned subfolders (`sqlite3-native/5.1.7/`, `ffmpeg/4.1.5/`, `git/v2.47.3-1/`)
- [ ] Upgrade simulation: manually place a binary in the old flat location; confirm it is moved (not re-downloaded) to the versioned folder on activation
- [ ] Version mismatch: corrupt or delete `version.txt` inside a versioned folder; confirm a re-download is triggered

### Retry behavior
- [ ] With no network, confirm each tool attempts the download exactly 2 times before giving up, then tries again on the next activation (no permanent block)
- [ ] With network restored after a failure, confirm a full download succeeds on the next activation without needing a manual trigger

### Tools Status UI
- [ ] With all tools missing, confirm "Download and Install" buttons are visible for all three tools
- [ ] Click a download button; confirm the button cycles through `Attempt 1 of 2`, `Download Failed` (on failure), back to `Download and Install`
- [ ] After a successful download, confirm the tool's mode switches from `builtin` → `auto` and severity updates without reloading
- [ ] Confirm the webview does NOT auto-open on startup; a toast appears only for a critical SQLite failure

### Tool status logging
- [ ] Confirm log line reports `native` when binary is present and mode is `auto`
- [ ] Confirm log line reports `available but using fallback` when binary exists but mode is `builtin`
- [ ] Confirm log line reports `missing and using fallback` when binary is absent and mode is `auto`

### Git binary (frontier-authentication)
- [ ] On a clean machine, Git binary downloads to `{globalStorage}/git/v2.47.3-1/`
- [ ] `git --version` executes correctly from the installed path
- [ ] With the native binary absent, isomorphic-git fallback handles basic sync operations
- [ ] Offline at startup: download is skipped cleanly with no hang; retry succeeds when back online

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -64,9 +64,8 @@ import { initializeAudioMerger } from "./utils/audioMerger";
 import { initializeAudioExtractor } from "./utils/audioExtractor";
 import { initializeAudioExporter } from "./exportHandler/audioExporter";
 import { checkTools, getUnavailableTools } from "./utils/toolsManager";
-import { initToolPreferences, setNativeGitAvailable, getGitToolMode, getSqliteToolMode } from "./utils/toolPreferences";
+import { initToolPreferences, setNativeGitAvailable, getGitToolMode, getSqliteToolMode, getAudioToolMode } from "./utils/toolPreferences";
 import { downloadFFmpeg } from "./utils/ffmpegManager";
-import { resetRetryCount } from "./utils/binaryIntegrityUtils";
 import { MissingToolsWarningProvider } from "./providers/MissingToolsWarning/MissingToolsWarningProvider";
 import { cleanupOrphanedProjectFiles } from "./utils/fileUtils";
 // markUserAsUpdatedInRemoteList is now called in performProjectUpdate before window reload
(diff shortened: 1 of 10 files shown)


BenjaminScholtens (approved):
works good

PR #916, opened 2026-04-24
USFM Importer Round-trip fix

### LINKED ISSUE #913: Fix USFM importer missing "importerType" issue
The importer is missing metadata field "importerType" and the Round-trip exporter is failing due to that. Quick fix

Fikitti:
Fixed the missing metadata information for the importer/exporter to work.

Got rid of the old unused USFM importer files that were messing with the structure of our file system, and the repo is now cleaner.

The importer was tested with several files both imported and exported correctly through Round-trip and USFM generate.


LeviXIII (commented):
I ran into an issue where the importer type was not set correctly on 2 of the files I used. It may be due to it being put into the OT Corpus Marker and grouping it with the eBible download that I had. However, if I imported it using USFM I would think that it should retain that type in the metadata. 

Exporting gave me an issue when trying to round-trip, but other options worked fine.

It also didn't work for an .sfm file that, while messed up, should still attempt to export.

LeviXIII (approved):

PR #920, opened 2026-04-25
Prevent requiredExtensions ratchet during Conductor pin-sync window

JonahBraun:
Prevent requiredExtensions from ratcheting up to a Conductor-pinned version during the mid-sync window when metadata.json on disk has no pinnedExtensions yet.

The ratchet suppression in `MetadataManager` read `pinnedExtensions` from `metadata.json` on disk. Per the pin-enforcement design (Scenario 2 in `codex-arch/2026-03-pin-enforcement-scenarios.md`), Frontier writes remote pins into the Conductor's workspace storage and **aborts the merge** before disk is updated. After the user reloads into the pinned profile, the running version matches the pin and `codex-editor` activates normally — but `metadata.json` on disk still has no `pinnedExtensions`. The ratchet guard saw no pin and bumped `requiredExtensions` up to the running (pinned) version. When the pin was later removed and the user reverted to Default, the inflated `requiredExtensions` blocked sync indefinitely.

The Conductor already exposes `codex.conductor.getEffectivePinnedExtensions` (admin intent → remote pins in storage → on-disk fallback) and is consumed by `failsPinMatchGate` and Frontier's sync gate. The ratchet now consults the same source.

## Changes

- Add `MetadataManager.getEffectivePinnedExtensions()` helper that queries the Conductor command and falls back to a caller-supplied disk pin map (covers older Codex shells without the command registered).
- `ensureExtensionVersionsRecorded`: use the helper for its short-circuit pin check.
- `updateExtensionVersions`: query the Conductor **inside** the `safeUpdateMetadata` queued callback so the pin snapshot reflects state after the per-workspace write-queue slot is acquired. Fixes a race where a Frontier write between the pre-queue snapshot and the actual write would still inflate `requiredExtensions`.
- Extend the "Pin-aware ratchet" test suite with a `vscode.commands.executeCommand` stub and 5 new cases:
  - Conductor-only pin suppresses ratchet when disk has no pins (bug repro).
  - Conductor throws → falls back to on-disk pins.
  - Conductor returns `{}` and disk has no pins → ratchet proceeds.
  - Conductor pins only frontier → codexEditor still ratchets.
  - Race guard: holds a slow `safeUpdateMetadata` open, enqueues `updateExtensionVersions` behind it, flips the Conductor stub mid-wait, asserts no ratchet.

Existing broken projects (already-inflated `requiredExtensions` on origin) are out of scope. Manual admin edit of `metadata.json` is the unblock; a separate `-pr` suffix fail-open in Frontier's sync gate is being considered for the PR-pin majority case.

## Test plan

- [ ] `npm run compile` clean (modulo the pre-existing `dugite` module-resolution error unrelated to this change).
- [ ] `npm run test` — new "Conductor-effective pins" sub-suite passes; existing Pin-aware ratchet cases still pass.
- [x] Manual E2E reproducing the user-reported flow:
  - Open project with no pins on Default profile (codex-editor v0.24.0). Confirm `requiredExtensions.codexEditor === "0.24.0"`.
  - Admin pins to v0.28.5, pushes.
  - User syncs → Frontier "sync paused" notification + Conductor "Reload Codex" prompt. `metadata.json` on disk unchanged.
  - Reload into the v0.28.5 profile. Inspect `metadata.json`: `requiredExtensions.codexEditor` is still \`"0.24.0"\` (pre-fix: inflated to \`"0.28.5"\`).
  - Sync completes; `pinnedExtensions` lands on disk; `requiredExtensions.codexEditor` still \`"0.24.0"\`.
  - Admin removes pins, user syncs, reloads to Default.
  - Subsequent sync passes: required \`0.24.0\` ≤ installed \`0.24.0\`.

### DIFF
--- a/src/test/suite/metadataManager.test.ts
+++ b/src/test/suite/metadataManager.test.ts
@@ -291,6 +291,173 @@ suite('MetadataManager Tests', () => {
             assert.strictEqual(versions.versions?.codexEditor, '0.22.90');       // suppressed
             assert.strictEqual(versions.versions?.frontierAuthentication, '0.4.25'); // ratcheted
         });
+
+        suite('Conductor-effective pins (mid-sync window)', () => {
+            const CONDUCTOR_CMD = 'codex.conductor.getEffectivePinnedExtensions';
+            type PinMap = Record<string, { version: string; url?: string; }> | undefined;
+            let originalExecuteCommand: typeof vscode.commands.executeCommand;
+            let conductorResponse: PinMap | (() => Promise<PinMap>) | 'throw';
+
+            setup(() => {
+                originalExecuteCommand = vscode.commands.executeCommand;
+                conductorResponse = undefined;
+                (vscode.commands as any).executeCommand = async (cmd: string, ...args: any[]) => {
+                    if (cmd === CONDUCTOR_CMD) {
+                        if (conductorResponse === 'throw') {
+                            throw new Error('Conductor unavailable (test stub)');
+                        }
+                        return typeof conductorResponse === 'function'
+                            ? await conductorResponse()
+                            : conductorResponse;
+                    }
+                    return originalExecuteCommand.call(vscode.commands, cmd, ...args);
+                };
+            });
+
+            teardown(() => {
+                (vscode.commands as any).executeCommand = originalExecuteCommand;
+            });
+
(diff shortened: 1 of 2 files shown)


BenjaminScholtens (approved):
looks good

PR #931, opened 2026-04-28
Vtt importer remove html fix

### LINKED ISSUE #927: Remove html enforcement from vtt importer
the html enforcement shouldn't be present in the vtt importer, since it is not a round-trip one

Fikitti:
Removed the HTML enforcement checkbox option from the VTT importer.

Added a feature too every file menu opened by three dots in the navigation, saying "HTML Enforcement" where user may enable or disable it. It works instantly when enabled/disabled :D

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "codex-editor-extension",
-    "version": "0.25.0",
+    "version": "0.26.0",
     "lockfileVersion": 3,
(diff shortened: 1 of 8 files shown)


LeviXIII (approved):
Made a couple of changes:

- Added a glow on enabling the switch
- Removed the save button and made it save on toggle
- Cancel made into Close button
- Removed shield icon

Issue #918, opened 2026-04-24
Create separate download folders per milestone


LeviXIII:
STEP 3:

This would allow for a folder for each chapter of a book of the Bible.

Issue #917, opened 2026-04-24
Allow for chapter and verse in filenames


LeviXIII:
STEP 2:

When saving audio files, the chapter and verse (if available) should be in the filename on export

PR #941, opened 2026-04-30
914 allow for exporting audio while using stream only

### LINKED ISSUE #914: Allow for exporting audio while using Stream Only
STEP 1:

People with large projects (lots of audio) need to be able to open the project in Stream Only and then export without having to reopen the project and download ALL of the audio.

Fikitti:
Added ability to export audio files for Stream Only. It uses similar approach as our playback feature.

Test Project: 941-allow-for-exporting-audio-while-using-stream-only

Tied to https://github.com/genesis-ai-dev/frontier-authentication/pull/31


Fikitti:
Resolves issues #914 #917 #918

TimRl (changes requested):
Batch downloading of the files should be implemented. It downloads extremely slow compared to other downloads in Codex.

LeviXIII (commented):
In order to unblock @Fikitti from the last part of the overall ticket (#918), I'll let this one through with the minor fixes as other tickets (#996, #997, #998). This PR requires a bump in Frontier Auth which will be done in a separate bump PR.

LeviXIII (approved):

PR #952, opened 2026-05-05
Add "Copy Error Details" action to sync failure toast

JonahBraun:
Add a "Copy Error Details" action to the sync-failure toast.

When sync fails, users now see an actionable button that copies a plaintext diagnostic report to the clipboard. If the underlying error is a `GitLabApiError` thrown by frontier-authentication, the report includes the HTTP status, URL, method, operation label, and full response body. Otherwise it falls back to the error message and stack. Pairs with frontier-authentication PR (`feat/api-error-details`) — both should ship together for the rich GitLab section to populate.

Toast w/ copy button looks like this:
Screenshot 2026-05-01 at 2 38 46 PM

## Changes
- Add `showSyncErrorWithCopy` and `formatSyncErrorReport` in `src/projectManager/utils/merge/index.ts`
- Replace the bare `vscode.window.showErrorMessage` in `stageAndCommitAllAndSync`'s catch with the new helper
- Duck-type `GitLabApiError` via `error.name` to avoid a runtime dep on frontier-authentication

## Test plan
- [x] Trigger a sync failure (e.g. revoke GitLab token) and verify the toast shows "Copy Error Details"
- [x] Click the button, paste into a scratch buffer, confirm the report contains status, URL, and response body
- [x] Trigger a non-GitLab failure (e.g. force a thrown Error in the catch path) and confirm the fallback report shows message + stack

### DIFF
--- a/src/projectManager/utils/merge/index.ts
+++ b/src/projectManager/utils/merge/index.ts
@@ -273,11 +273,91 @@ export async function stageAndCommitAllAndSync(
         return syncResult;
     } catch (error) {
         console.error("Failed to commit and sync changes:", error);
-        vscode.window.showErrorMessage(
-            `Sync failed. Please try again or contact support if the problem persists.`
-        );
+        // Fire-and-forget: showErrorMessage with an action only resolves when
+        // the user dismisses/clicks the toast, and we must not block the sync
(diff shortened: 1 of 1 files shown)


LeviXIII (approved):
New error copied message shows up correctly.

PR #959, opened 2026-05-06
Vsix minification

### LINKED ISSUE #859: Minify the vsix

TimRl:
## Summary

This PR reduces the size of the generated VSIX by tightening what we package and making the production extension build smaller.

### What changed

- Uses the production webpack package build for VSIX output so `extension.js` is minified.
- Cleans `out/` before the final package build so stale test or chunk output is not accidentally included.
- Builds only the extension webpack config for the final package output, while still running the full compile/test pipeline before packaging.
- Excludes source maps from the published VSIX.
- Excludes `.vscode-test/` and `package-lock.json` from the VSIX.
- Avoids broad asset/package allowlists like `!**/*.svg` and `!package*.json` that can accidentally pull large unrelated files into the package.
- Keeps `dugite` external while packaging only the small JS wrapper metadata needed at runtime, not dugite’s embedded Git binaries.
- Keeps copied runtime assets, such as `fts5-sql-bundle` and codicons, under `out/node_modules` so `.vscodeignore` can safely include a narrow runtime output path.
- Restricts `pdf-parse` bundling to its default `pdf.js` version instead of pulling in every bundled pdf.js version.
- Adds comments documenting the VSIX packaging contract so future dependency additions do not accidentally bloat or break the package.

## Why

The previous VSIX packaging could include unnecessary production artifacts such as source maps, test installs, lockfiles, and stale build output. This made the VSIX larger than needed and made future dependency additions easy to get wrong.

The new pattern is:

- Bundle what can be bundled.
- Externalize only what needs real runtime paths.
- Copy required external runtime files into `out/node_modules`.
- Allowlist narrow output paths in `.vscodeignore`.
- Keep broad root dependency folders out of the VSIX.

## Test Plan

- Verified the extension builds successfull (vsix)

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -18,8 +18,6 @@
 # Runtime dependencies
 !node_modules/dugite/build/**
 !node_modules/dugite/package.json
-!node_modules/@vscode/codicons/dist/**
-!node_modules/vscode-uri
 
 # Documentation
 !CHANGELOG.md
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -2,30 +2,31 @@
 *
 */**
 
-# Allowlist what we need (only extension output, NOT test bundles)
+# Extension output (JS bundles only)
 !out/*.js
-!out/*.js.map
 !out/node_modules/**
+
+# Static assets
 !assets/**
+!src/assets/reset.css
+!src/assets/bible_data_with_vref_keys.json.gz
 
-# Exclude test output and source maps from VSIX
-out/test/**
-out/test-runner/**
-out/**/*.map
-.vscode-test/**
-# !servers/**
-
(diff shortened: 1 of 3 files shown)


BenjaminScholtens:
Screenshot 2026-05-07 at 10 32 52 AM
the above build done with the github action created a failing extension.

BenjaminScholtens (dismissed):
Fix build done with GitHub action.

TimRl:
> Fix build done with GitHub action.

Okay, that's a fail on my part. Thanks for that @BenjaminScholtens.

It's been fixed now.

LeviXIII (approved):
Creates a package half the size of the original. Runs like normal without any noticeable bugs.

PR #960, opened 2026-05-07
951 we need the option for a project admin to clean up a source and target file by hiding cells that dont need to be imported

### LINKED ISSUE #951: We need the option for a project admin to clean up a source and target file by hiding cells that don't need to be imported.
Image

Perhaps we should add this to source editing mode? 

We just need an option in the schema to hide the cells. They are still there but don't need to be translated.

Fikitti:
## Summary
Closes #951

Adds a way for a project admin to hide cells that don't need to be translated. In source editing mode, each cell now has an eye button next to the merge button. Clicking it hides the cell in normal mode. The cell is not deleted — it stays in the file with a `hidden` flag in its metadata, and can be shown again at any time.

## Changes
- New eye button on each cell in source editing mode to hide or show that cell
- Hidden cells stay in the `.source`/`.codex` files (marked with `data.hidden`) but do not appear in normal editing mode
- Hiding a cell in the source file also hides it in the matching target file (and the other way around)
- In source editing mode, hidden cells stay visible but dimmed so the admin can find and un-hide them
- Hidden cells are skipped by AI translation / autocomplete
- Hidden cells are left out of exports (USFM, HTML, subtitles, audio, delimited)
- Round-trip exports (TMX/XLIFF, DOCX, spreadsheet) keep the original untouched text for hidden cells instead of writing a translation
- Hide/show actions are recorded in the cell's edit history so they sync correctly between project members
- Also fixes TMS round-trip export bugs: numeric unit IDs now match after import, and translations are written into the `<tuv>` that matches the target language instead of always the second one

### Hiding and showing cells (verified by tests + code review)
- [x] Toggling a cell writes `hidden` to cell metadata and records an edit entry (syncs via merge resolver's generic data-path handling)
- [x] Hidden cells are filtered out of the normal editor view but still shown (dimmed) in source editing mode
- [x] The paired file gets the same hidden flag when a cell is toggled
- [x] Manual check: eye button appears only in source editing mode and toggles correctly in the running extension

### Exports (verified by tests)
- [x] Hidden cells are excluded from SRT/VTT subtitle exports
- [x] Hidden cells are excluded from active-cell filtering used by USFM/HTML/audio exports
- [x] TMX/XLIFF round-trip export keeps the original target text for hidden, deleted, and merged cells
- [x] Spreadsheet and DOCX exports keep original content for hidden cells

### It doesn't do the other thing
- [x] Hidden cells are not deleted — they remain in the file JSON
- [x] AI autocomplete does not translate hidden cells
- [x] Merged and deleted cell behavior is unchanged (existing tests still pass)


LeviXIII (approved):
Just made sure to add the edit history in the metadata for each cell that gets hidden or revealed.

PR #961, opened 2026-05-07
Add translation influence inspection

dadukhankevin:
## Summary
- Adds a similar-wording inspection feature that surfaces which prior translations influenced the current cell's draft
- Introduces `SimilarWordingDialog` UI in the Codex Cell Editor and a new `similarWordingInspection.ts` analysis module
- Extends search internals (`sqliteIndex`, `contextBranchingSearch`, `searchManager`) and message handling to support the inspection flow

## Test plan
- [ ] Open a translated cell and trigger the similar-wording inspection — verify the dialog lists relevant influencing cells
- [ ] Confirm search results are scoped/ranked correctly across books and chapters
- [ ] Verify no regressions in existing CodexCellEditor message handling


LeviXIII:
## What this branch adds

A new "Why this translation?" inspection dialog in `CodexCellEditor` that takes the current target cell content, uses the SBS (context branching) search algorithm to find target-side wording matches across the project, and shows which fragments of the translation appear in other already-translated cells. Plumbing-wise it adds:

- A new `searchScope: "source" | "target" | "both"` option threaded through `BaseSearchAlgorithm` → `searchManager` → `ContextBranchingSearchAlgorithm` → `SQLiteIndexManager.searchCompleteTranslationPairs[WithValidation]`.
- A new `excludeCellIds` option on the search base.
- A new `searchWithBranchMatches` method on SBS that returns the *covered substring* per match.
- A new module `similarWordingInspection.ts` that maps SBS results into chunks/occurrences.
- A new extension-host handler `requestSimilarWordingInspection` and two new return messages (`similarWordingInspectionResult` / `similarWordingInspectionError`).
- A new webview component `SimilarWordingDialog.tsx`, plus button + state/handlers in `TextCellEditor.tsx`, `CellList.tsx`, `CodexCellEditor.tsx`.
- Two **non‑additive** internal changes piggybacked in the same commit (see "Risk #1" below):
  - SBS candidate scoring formula changed.
  - Punctuation regex in `cleanQuery` / SBS tokenizer changed from `[^\w\s]` → `[^\p{L}\p{N}\p{M}\s]/u`.

## Testing checklist

### A. New "Why this translation?" feature (happy path)

- [x] Open a `.codex` file, click into a target cell that has translated content. A new "Why this translation?" button (with `AlertCircle` icon) appears in the editor toolbar next to existing buttons (`webviews/codex-webviews/src/CodexCellEditor/TextCellEditor.tsx` ~2195).
- [x] Button is **hidden** on a `.source` (read‑only) cell (`isSourceText` guard).
- [x] Button is **disabled** when the editor content is empty / only whitespace/HTML.
- [x] Clicking opens the `SimilarWordingDialog` showing "Searching..." state, then either chunk results, a "No matching translated examples were found" empty state, or an error block.
- [x] Header shows the current cell's plain text with chunk fragments rendered as highlighted clickable buttons.
- [x] Clicking a chunk filters to just that chunk; "Show all chunks" button restores the list.
- [x] Each occurrence card shows: cell label (via `getCellDisplayLabel`), validated/not-validated badge, target snippet, optional source snippet underneath in muted text.

### B. Wiring to other features

- [x] Click the magnifying‑glass on a chunk → Parallel Passages sidebar focuses and runs a search for the chunk text. (Calls `expandSearchToAllFiles`, which is wired in `codexCellEditorMessagehandling.ts` line ~3558 and uses `search-passages-sidebar.focus` + `setPendingSearch`.)
- [x] Click the pin icon on an occurrence → Parallel Passages pins that cell. (Calls `parallelPassages.pinCellById`, registered in `src/providers/registerProviders.ts:77`.)
- [x] Closing the dialog (X / Esc / outside click) clears state; reopening on the same cell reissues the request and resets `selectedChunkIndex`.

### C. Edge cases

- [x] Target content **shorter than `DEFAULT_MIN_TOKENS` (3 tokens)** → dialog opens with "No translated text to inspect." (handled by the early return in `inspectSimilarWording`).
- [ ] Target content with only HTML / `&nbsp;` → button stays disabled (the webview's `hasInspectableEditorContent` strips tags & entities before checking length).
- [ ] **Non‑Latin script** target content (e.g. Devanagari, Cyrillic, Hebrew, Arabic) → token regex `\p{L}\p{N}\p{M}` should now correctly tokenize. Verify the dialog finds chunks at all for such scripts — this is the main reason the punctuation regex was changed.
- [ ] Very **long** target cell (>18 tokens) → `createInitialBranches` should split the query into windowed branches (window 14, stride 8, capped at 12). Verify it still returns chunks rather than nothing.
- [ ] Cell whose `cellId` matches itself via duplicate translation in another file → should be excluded (`excludeCellIds: [args.cellId]`).
- [ ] Index not yet ready (project just opened) → handler throws "Search index is not ready." which surfaces as `similarWordingInspectionError` — verify the dialog renders the red error banner instead of crashing.
- [ ] Index manager has been closed (e.g., after project swap) — should also surface as an error message, not a silent freeze.

### D. Regression: SBS-driven features (smart edits / predictive translation)

The SBS algorithm is also the configured default for `getTranslationPairsFromSourceCellQuery`, which feeds **few-shot examples to the LLM**. The scoring formula changed, so do an A/B sanity check:

- [ ] Trigger an LLM completion (predictive translation) on a few cells in an existing project; confirm the suggestion still feels reasonable and isn't obviously degraded (or is improved).
- [ ] Run the existing test suite: `npm run test` and especially `src/test/suite/validatedOnlyExamples.test.ts`. The mock there only checks `onlyValidated` is forwarded; that should still pass, but re-confirm.
- [ ] If you have a project with non-Latin script translations, sanity-check that few-shot retrieval still returns sensible matches (the tokenization regex change affects this code path too).

### E. Regression: existing FTS search behavior

Both `searchCompleteTranslationPairs` and `searchCompleteTranslationPairsWithValidation` got new optional `searchScope` parameters and slightly different SQL `line` / `uri` expressions. Default behavior is preserved (when `searchScope` is undefined, it falls back to the old `searchSourceOnly` boolean).

- [ ] In‑tab search bar (`Ctrl/Cmd+F` → FloatingSearchBar) inside a `.codex` file: confirm match counts and navigation still work.
- [ ] "Expand to all files" / Parallel Passages search: confirm results render with correct file paths and line numbers (the SQL `uri`/`line` expressions are now scope-specific instead of always `COALESCE`).
- [ ] Searches containing **punctuation** (`don't`, `Jesus, Lord`, `1:23`) — the cleaned query no longer collapses inside-word marks like apostrophes the same way; verify expected hits still come back.

### F. Build / type sanity

- [ ] `npm run compile` and `npm run lint` from repo root succeed.
- [ ] `pnpm run build:CodexCellEditor` (or `build:all`) inside `webviews/codex-webviews/` succeeds — the new types in `types/index.d.ts` are imported by both extension and webview.
- [ ] `pnpm run test` inside `webviews/codex-webviews/` passes (vitest).
- [ ] No new console errors when opening `.codex` files (the new `useMessageHandler("codexCellEditor-similarWordingInspection", …)` registers another listener; verify it doesn't double-register on hot reload).

### G. Manual UI polish

- [ ] In a narrow editor pane, the button collapses to icon-only (`hidden sm:inline` on the label). Verify it doesn't overflow the toolbar.
- [ ] Dialog at small heights — the "Current Translation" panel scrolls (`max-h-32 overflow-y-auto`) and the chunks list scrolls beneath it.
- [ ] Keyboard: dialog is dismissible with Esc; chunk buttons are focusable.

LeviXIII (approved):

PR #965, opened 2026-05-11
956 verse ranges remain missing with merge verse recovery

### LINKED ISSUE #956: Verse Ranges remain missing after closing issue 848
It doesn't seem to be syncing the change correctly.

LeviXIII:
Same as 956 but with a command to recover merged verses that may have been left behind.

Closes #956

**PROJECT TO TEST WITH:** `956-verse-ranges-remain-missing-with-merge-verse-recovery`

This branch fixes verse-range cells that go missing or get doubled across the verse-range migration + git sync, and adds a one-off recovery command for projects that were already damaged. Six commits, layered:

### Happy path
- [ ] Run the command from the palette.
- [ ] The pre-run warning dialog appears with the expected copy.
- [ ] Click **Cancel** → the dialog closes cleanly; **no second cancel prompt** appears (regression check for the double-cancel fix).
- [ ] Run the command again and click **Apply** (or whatever the confirm action is).
- [ ] Progress notification shows file-by-file scan.
- [ ] After it finishes, open an affected `.codex` file and confirm:
  - Parent verse-range cell contains the concatenated content of all its children.
  - Children of the range are hidden from the editor (soft-deleted, not removed).
  - `cellLabel` on the parent matches the range (e.g. `21-24`).
  - The range parent sits **after** its chapter milestone and **in verse order**.
  - Paratexts (e.g. section headings) appear **above** their parent verse.

### Idempotency
- [ ] Re-run the command. Confirm it reports nothing to migrate / the files are byte-identical to the previous run.
- [ ] Run on a non-scripture notebook (notes/glossary). Confirm it is a no-op.

## Multi-user sync survival (the newest fix) <- MOST IMPORTANT TO TEST AS THE OTHER ITEMS WERE TESTED ALREADY

> Goal: prove the migration's output persists across multiple sync round-trips from different users.

- [ ] **User A**: run the migration locally on a notebook with a parent that has **3+ merged children**, then push.
- [ ] **User B** (whose clone has *never* run the migration): pull. Confirm in User B's editor:
  - Parent value contains every merged child's content (not just the first one).
  - All merged children are soft-deleted on User B's side.
  - `metadata.data.mergedChildIds` on the parent lists every child.
- [ ] **User B** makes an unrelated edit on a different cell in the same file and pushes.
- [ ] **User A** pulls. Confirm the parent value, soft-deletes, and `mergedChildIds` are all still intact (no truncation back to just the first child).
- [ ] Repeat the A→B→A round-trip **2 more times**. State must remain intact every round.
- [ ] On either clone, re-run the migration. Confirm it is a no-op (no doubled content, no new edits).

## "Codex: Recover Missing Merged Children" command

> Use on a project that already had the legacy broken state, OR construct one by hand-editing a parent cell to drop a soft-deleted child's content out of `parent.value`. 

**You can check Obadiah and Philemon for changes here.**

- [ ] Run the command.
- [ ] Scan finishes; warning dialog reports `N parent verse cells across M files`.
- [ ] Click **Cancel** → no file is modified (verify via `git status` or diff).
- [ ] Run again and click **Apply**.
- [ ] Affected parents now contain the missing children's content.
- [ ] Re-run the command. Confirm it reports **"No missing merged children detected."** (idempotent.)

### Recovery correctness (no false positives)
- [ ] **Footnote child**: parent whose soft-deleted child contains a `<sup class="footnote-marker" data-footnote="...">` element with HTML inside the attribute value. The child's content is already in the parent (under escaped form). → recovery must **NOT** re-append it.
- [ ] **Lightly-edited parent** (one word inserted into the existing merged content): recovery must **NOT** re-append the child (fuzzy bigram match).
- [ ] **Digit-vs-spelled-number variant** (e.g. parent has "five hundred" where the child had "500"): recovery must **NOT** re-append.

### Recovery correctness (true positives)
- [ ] **Genuinely missing child** whose content is wholly absent from the parent: recovery **must** re-append.
- [ ] Mixed file: one parent has a real loss, sibling parents are fine. Only the affected parent gets an edit; the others are untouched.

## Regression checks

- [ ] Open a non-verse-range notebook (e.g. plain prose). Confirm the migration command and the recovery command are no-ops on it.
- [ ] Hand-edit a `cellLabel` on a verse-range cell. Run the migration. Confirm the user's label is preserved (not overwritten by the autofill).
- [ ] Run the project's normal sync (`stageAndCommitAllAndSync`) with no migration-related changes pending. Confirm sync behaves normally and the new error-toast "Copy Error Details" action still works when sync fails.


Luke-Bilhorn (approved):
Matt said this has been thoroughly reviewed!

PR #967, opened 2026-05-12
Add character audio export functionality and update audio export options

### LINKED ISSUE #966: Export audio consolidated by character labels
User needs to export audio from a project but consolidated by character labels.

In a subtitle file, the character labels will be something like "JESUS" or "MARY MAGDALENE", etc. 

- [x] The final export should have one audio file per character. 
- [x] Each exported audio file should be the full length of the episode.
- [x] The audio should be simply silence until the specific character speaks.
- [x] The exported files should be labelled with `[filename/episode name]_[lang-code]_[characterlabel].wav`.
- [x] This should be an option in the existing export flow - under "Audio Export Options"

ryderwishart:
- Introduced a new `characterAudioExporter` module for consolidating audio exports by character.
- Updated `exportHandler` to support new options for consolidating audio and specifying audio formats.
- Enhanced the project export view to include a preview feature for character audio.
- Added new entries to `.gitignore` for Glance memory shards.
- Made several functions in `audioExporter` module publicly accessible for use in the new character audio export logic.


LeviXIII (approved):
One thing I encountered when exporting as flac was that if opening in a media player, the longer files don't play at their respective places. The files were meant to be dropped into a DAW, so playing them this way isn't the intent or purpose. However, it should be noted that .wav files will work best when trying to just play the media.

PR #972, opened 2026-05-13
fix(package): update package script to use Node's fs.rmSync for bette…

TimRl:
…r cross-platform compatibility


LeviXIII (approved):
Helps Windows users to package.

PR #974, opened 2026-05-18
Milestone Editing

### LINKED ISSUE #936: Make Milestones Editable
Milestones themselves should be editable. Expand the UI from 867 (adding subdivisions) to include editing milestones as well. This includes making them editable only on the source side, and not on the target side, but allowing for translations of saved names on the target side. Milestone cells themselves should be moved. 

To test, 
- make sure that nothing breaks when you change the milestones around on an existing project
- make sure that switching the ordering of milestones is disallowed
- make sure deletion of milestones is accepted
- make sure that numbering of milestones remains accurate. 

Also, add an "Edit milestones" permission in settings that is turned off by default; this way we can hide it from users that really should not be editing their milestones, like most BT users, but those who actually need it can find it in the settings. This would be something we could even just turn on by default for pinned projects of subtitle users.

Luke-Bilhorn:
Expands the subdivision editing PR to include milestone editing, with associated tests

### 1. Setup & Migration
*Deleted; this migration was not for this Milestone PR*

### 2. Interface Settings panel
* [x] Open **Interface Settings** → "Pagination & Subdivisions" section renders.
* [x] **Cells per page**: enter a valid value (e.g. 25) → persists; reopening editor repaginates milestones to 25/page.
* [x] **Cells per page** clamping: enter `3` → clamps to 5 on commit; enter `250` → clamps to 200; enter non-numeric → reverts to last valid; Enter key commits.
* [x] **Maximum subdivision length** toggle OFF → stored as 0 (resolver falls back to `cellsPerPage` as the threshold).
* [x] **Maximum subdivision length** toggle ON → number input appears, pre-filled (~`cellsPerPage × 2`); clamps to 1–5000; invalid reverts.
* [x] **Enable Milestone Placement Editing** toggle ON/OFF → change is reflected live in an open editor's accordion (structural controls appear/disappear without reload).
* [x] Change a setting from the **VS Code Settings UI** (not the panel) → Interface Settings panel and open editors stay in sync.
* [x] **`useSubdivisionNumberLabels`** is currently hidden in the Interface Settings panel — verify it can still be toggled via VS Code `settings.json` and takes effect (see Display section).

### 3. Milestone display & pagination
* [x] Milestone accordion lists all milestones with correct labels ("Book Chapter", e.g. "Isaiah 1") and text/audio progress indicators.
* [x] A milestone with **no custom breaks** paginates arithmetically into `cellsPerPage`-sized subdivisions (default 50 → "1–50", "51–100", …) — matches legacy pagination exactly.
* [x] A milestone shorter than `cellsPerPage` shows a single subdivision (no chunking).
* [x] With `maxSubdivisionLength` set (e.g. 200) and a long stretch between breaks (e.g. 300 cells): the stretch sub-chunks into `cellsPerPage`-sized pages; a stretch **shorter** than the threshold stays unsplit.
* [x] Clicking a subdivision navigates to / loads that page of cells.
* [x] Navigation header reflects the correct milestone/subdivision while paging.

### 4. Subdivision label preference (`useSubdivisionNumberLabels`)
* [x] Default (false): a **named** subdivision shows its name, with the numeric range as a muted suffix.
* [x] Set true: subdivisions always show the numeric range as the primary label even when a name exists.
* [x] Toggling the setting updates open editors **live** (no reload) and does not repaginate.

### 5. Settings/gear gating & source-only rules
* [x] With placement editing OFF: the gear toggle reveals **only** rename pencils (milestone + subdivision); no add/remove/promote/demote controls.
* [x] With placement editing ON: the gear reveals add-milestone, remove (trash), promote (↑), demote (↓) controls.
* [x] Gear toggle hides all edit affordances when off; closing & reopening the accordion collapses settings mode back to default.
* [x] On a **target** (`.codex`) document: "Add break", "Add milestone", delete-subdivision, promote/demote, and reset controls are **not shown**; milestone + subdivision rename pencils still appear.
* [x] Provider rejects any structural/subdivision **write** that originates from a target document (defense-in-depth — even if UI is bypassed, a warning is shown and nothing changes).

### 6. Milestone rename
* [x] Click the milestone pencil → inline input replaces the title, text auto-selected.
* [x] Save with a valid changed value → label updates immediately (optimistic), persists, and mirrors to the target milestone.
* [x] Save button disabled when value is empty, whitespace-only, or unchanged; whitespace is trimmed on save.
* [x] Enter saves, Escape reverts (revert sends no message).
* [x] Milestone rename works on **both** source and target documents (renaming is allowed on target; it does **not** mirror back to source).
* [x] Source rename **overwrites** a target-side rename (source is authoritative — no target "stickiness").

### 7. Subdivision rename
* [x] Rename pencil appears only for subdivisions that carry a key (custom or auto-chunk with a cell anchor).
* [x] Saving a name posts `updateMilestoneSubdivisionName`; the name displays and the numeric range stays visible.
* [x] Naming an **auto-chunk** on the source promotes it into a persistent custom placement.
* [x] Naming the **implicit first** subdivision does **not** create a placement (stored under `__start__`).
* [x] Clearing a name (empty string) leaves the placement intact (does not delete the break).
* [x] Unchanged name → no message posted.
* [x] **Focus is retained** in the rename input across parent re-renders (regression: focus loss was fixed) — type, let a re-render happen, keep typing.
* [x] Target-side rename never creates a placement; target name overrides source name locally, but an empty target override falls back to the source-mirrored name.

### 8. Add subdivision break (source only)
* [x] "Add break…" button shows on source when the milestone has ≥2 cells; hidden when the milestone has only 1 cell.
* [x] Enter a valid cell number (2..last) → new break created at the Nth root cell; existing source names preserved.
* [x] Out-of-range or empty input → inline error ("Enter a number between 2 and N." / "This milestone is too short to split."), nothing posted.
* [x] Re-adding a break at the same cell is **idempotent** (no duplicate).
* [x] Cancel/Escape closes the form without posting.
* [x] The new break mirrors to the paired target.

### 9. Reset subdivisions (source only)
* [x] "Reset to default breaks" appears only when custom breaks exist.
* [x] Reset requires **two clicks** (arm-then-confirm, auto-disarms after ~3s); second click clears all custom breaks and restores arithmetic pagination.
* [x] Reset mirrors to the target.

### 10. Add milestone at cell (split) — placement editing ON, source only
* [x] "Add milestone…" appears only with the feature on, source, settings mode.
* [x] Enter a valid cell number (≥2) → a new milestone is inserted there, splitting the original; placements before/after the boundary are redistributed correctly; a placement exactly at the boundary becomes the new milestone's first-subdivision name.
* [x] New milestone with no inherited name gets the **"New milestone"** placeholder label.
* [x] cellNumber < 2 is rejected (no-op).
* [x] The insertion + redistributed subdivisions **mirror to the target** by UUID.
* [x] After the split, the editor repositions correctly (see Position restoration).

### 11. Promote subdivision → milestone (placement editing ON, source only)
* [x] Promote (↑) appears on custom subdivisions that are **not** the first subdivision.
* [x] Single click promotes: the subdivision's name becomes the new milestone's label; the source placement at that anchor is removed (it's now a milestone).
* [x] An **unnamed** promoted subdivision yields the "New milestone" placeholder.
* [x] The first/implicit subdivision cannot be promoted (control hidden; provider refuses).
* [x] Promotion mirrors to the target; if the translator had **locally named** that seam on the target, the target milestone keeps the local name.

### 12. Demote milestone → subdivision (placement editing ON, source only)
* [x] Demote (↓) appears on every milestone **except the first**.
* [x] **Single click** demotes (no confirm): the milestone merges into the previous milestone, and the boundary is **preserved** as a custom subdivision carrying the demoted milestone's label.
* [x] First milestone never shows demote (or remove).
* [x] Demotion mirrors the merge + boundary onto the target; target translator's subdivision names from the removed milestone fold onto the survivor.

### 13. Remove milestone (placement editing ON, source only)
* [x] Remove (trash) appears on every milestone except the first (first shows a disabled/ghost trash).
* [x] Remove requires **two clicks** (arm-then-confirm within ~3s, tooltip explains "content merges into the previous milestone").
* [x] After removal: content cells merge into the previous milestone; subdivision breaks inside the removed milestone are lifted onto the survivor; the boundary itself is **not** preserved (contrast with demote).
* [x] First milestone cannot be removed (UI disabled; provider refuses with a warning).
* [x] Removal mirrors the soft-delete onto the target by UUID.

### 14. Delete a single subdivision break (source only)
* [x] Trash icon appears only on **custom** subdivisions (auto/first show a disabled ghost icon).
* [x] Single click removes just that break (posts the filtered placement list); pagination recomputes; mirrors to target.

### 15. Source → target mirroring (cross-cutting)
* [x] For each operation above, open the paired source and target side-by-side and confirm the target reflects: renames, added/removed breaks, splits, merges, promotes, demotes.
* [x] Mirroring is **by UUID** and is **skipped (no target change, no crash)** when the source and target milestone roots have diverged (e.g. target edited independently). Verify a divergent case doesn't corrupt or delete misaligned target content.
* [x] Target document does **not** need an open webview for the mirror to apply (background open path) — verify after closing the target, editing source, then reopening target.

### 16. Text editing ↔ milestone interactions (via `TextCellEditor`)
*Matt and I decided we did not need to test it rigorously—it works.*
* [x] *Start editing a cell's text (dirty state) → in the milestone accordion, subdivision rows become **dimmed / not-clickable** and the **"Save changes first to change section"** banner appears.
* [x] With unsaved changes, click a subdivision (or another milestone) → navigation is **blocked** (no page change); confirm no `requestCellsForMilestone` fires.
* [x] Save the edit → subdivision/milestone navigation **re-enables** and works normally.
* [x] After saving a text edit inside a subdivision, the editor **stays on the same milestone/subsection** (pagination cache rebuilds without throwing you to a different page) and progress indicators update.
* [x] Edit & save text in a cell within a **custom-broken** milestone → the custom breaks and their names are preserved (cache invalidation rebuilds subdivisions, doesn't reset them to arithmetic).
* [x] **Add a new content cell** via the editor → the root-cell count grows: "Add break at cell…" max/range reflects the new count, existing cell-ID-anchored breaks stay put, and a trailing auto-subdivision extends/appears as expected.
* [x] **Delete (soft-delete) a content cell** that is a subdivision's anchor (`startCellId`) → the stale break is silently pruned and pagination falls back gracefully (no empty/broken subsection).
* [x] Confirm **milestone cells themselves are not editable as normal text** in `TextCellEditor` (no value/edit affordances for `CodexCellTypes.MILESTONE`); their label is changed only via the accordion rename.
* [x] On a **target** document: editing/saving cell text does not alter source milestone structure; the unsaved-changes gate applies independently per open document.
* [x] Edge case — hold **unsaved text edits**, then trigger a forced milestone refresh (e.g. a source-side structural edit mirrors to this target) → verify the unsaved edit is handled sanely (not silently lost without warning) and the view repositions correctly.*

### 17. Staying in the right place after edits and navigation
* [x] After you **promote, demote, add, or remove a milestone**, the editor automatically jumps to and shows the correct section — you're never left staring at the wrong chapter, the previous spot, or a blank page.
* *deleted*
* [x] When you **first open a file**, it opens to the expected section/page.
* [x] After an edit moves you to a new section, the view **stays put** — it doesn't flicker, bounce back to the old spot, or freeze.

### 18. Persistence & reload
* [x] After each edit type, **save and reopen** the file (and the project): changes persist on disk in the milestone cell's `metadata.data` (subdivisions / subdivisionNames / subdivisionNamesFromSource), and the milestone/subdivision structure renders identically.
* [ ] Edits survive a **git sync / 3-way merge** without losing the milestone label (initial-import edit anchors the label).
* [x] `cellsPerPage` / `maxSubdivisionLength` changes invalidate the pagination cache and repaginate on next view.

### 19. Regression / smoke (non-milestone functionality unaffected)
* *Deleted; there is no migration*
* [x] Cells with `parentId` (children), paratext, and soft-deleted cells are correctly **excluded** from root-content-cell counting (so break-at-cell numbering matches what the user sees).
* [x] A document where all milestones collapse to one "virtual" milestone (no real milestone cells) still paginates and renders.


Luke-Bilhorn:
When testing, make sure to go into interface settings, and enable milestone editing. Otherwise, it is disabled.

LeviXIII (commented):
Pagination & Subdivisions Feedback
1. Remove the extra space and period from "Cells per page"
2. Change wording as "this" isn't on the screen yet (the input only appears after enabling). Maybe you can say "...subdivisions up to the set length".

Image

Source Controls
1. Just need some tooltips to explain what the arrow does. I thought it meant to reorder, but I see that it just moves the selection between being a break and a milestone. Maybe find another icon to suggest exchange?
2. I would remove the ellipsis off of the buttons and add captials for Add Break and Add Milestone

Image

Webview
1. When adding breaks, the line numbers aren't accurate. I think they are still lining up by the overall cell count set in the settings.
2. Breaks and Milestone changes aren't reflected in the target webview (for audio files at least).
3. After making breaks and/or milestones, if I close the dropdown and then open it again, the delete for the first selection of break/milestones is enabled (it should be disabled)

Image

LeviXIII (commented):
When adding a break, if I edit the cell in the target, it seems to save, but doesn't appear in the cell until I reload the webview. Any edits to the text after that appear as they should.

Luke-Bilhorn:
For 10.2, we have equivalent behavior. We don't allow placement at the actual milestone boundary, because the user already has a subdivision there that they can rename.

Luke-Bilhorn:
Todo: Milestone deletion is not available.

They can demote it and then delete the subdivision, though, if they need to.

Luke-Bilhorn:
image

Luke-Bilhorn:
Checking off all of section 16 because the cell editor does work

Luke-Bilhorn:
17.2 Should be deleted

Luke-Bilhorn:
Section 1 was not necessary, confirmed

Luke-Bilhorn:
That paratext bug we noticed earlier happens on main also. I also see that adding paratext cells after adding child cells pushes those child cell numbers up.

Luke-Bilhorn:
As of now, everything is accounted for, except testing 3-way sync.

LeviXIII (approved):
Everything seems to be working now. Ran regression tests and tested syncing. Data comes in correctly.

PR #986, opened 2026-05-26
984 audio countdown option needs to be local instead of project setting

### LINKED ISSUE #984: Audio countdown option needs to be local instead of project setting
It's set to the entire project (localProjectSettings), so everyone will use it. We want it to be per computer (user). Here are the keys to use:

recordingCountdownSeconds
autoRecordOnMicClick

Image

After the changes, User A should be able to turn off recordingCountdownSeconds (or adjust it to 1, 2, or 3 seconds) and then sync. User B should not take in User A's change for neither recordingCountdownSeconds nor autoRecordOnMicClick. You shouldn't see the keys in localProjectSettings anymore, but in some sort of Global Settings.

Luke-Bilhorn:
Recording Countdown Seconds, Auto Record On Mic Click, and also Auto Download Audio on Open are all per-project-per-user now. 

Matt and I have tested all of these things and found them to be per-project-per-user, across syncs.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -352,9 +352,10 @@ export async function activate(context: vscode.ExtensionContext) {
 
     initToolPreferences(context);
 
-    // Per-user audio recording preferences (autoRecordOnMicClick,
-    // recordingCountdownSeconds) live in globalState. Initialize before any
-    // provider reads them; the per-workspace migration runs lazily below.
+    // Per-user audio preferences (autoDownloadAudioOnOpen,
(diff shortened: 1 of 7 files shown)


LeviXIII (commented):
User settings are now placed into the following spots for reference:

Mac: ~/Library/Application Support/<HostApp>/User/globalStorage/state.vscdb
Windows: %APPDATA%\<HostApp>\User\globalStorage\state.vscdb
Linux: ~/.config/<HostApp>/User/globalStorage/state.vscdb

LeviXIII (approved):

PR #992, opened 2026-05-27
Fix lost files in sync

### LINKED ISSUE #991: Lost files from Bubble in discord chat
When I sync a project, the project gets merged, and after the merge some files are lost (possibly because it has been a long time since my last sync). One example: in the gn-ji-an project, the files added in fbb0c4af80554630c2b1a1dc9f7b802954972c68 are lost after these two branches, b3b0945c and fbb0c4af. You can see that the parent of b3b0945c is 228171d3 and fbb0c4af , not fbb0c4af. What caused this? (I later retrieved the lost files using git commands, in the latest commit.) 

Image

Image

TimRl:
Fixes #991


dadukhankevin (approved):
cool code

PR #993, opened 2026-05-27
fix(audio): restore recording count badge on the History button (#981)

### LINKED ISSUE #981: Put number of audio recordings badge back on history button
The history button used to have the number of audio files available as a tag/badge. We need to put it back so that the user is aware that there is audio available instead of just showing the record button. We should also include soft-deleted files in that number since they can be restored.

Image

Image

Luke-Bilhorn:
The History button used to show a count badge so users could see at a glance how many recordings were stored for the cell — including soft-deleted ones that can still be restored. The badge was removed in 20237e35 along with other count UI; this re-adds only the History button badge, keeping the change minimal.

- Track audioHistoryCount in TextCellEditor from the existing audioHistoryReceived response (history.length already includes soft-deleted entries via getAttachmentHistory).
- Reset hasAudioHistory and audioHistoryCount on cell switch to avoid briefly rendering a stale count from the prior cell.
- Render a theme-aware count badge in both History button render sites: AudioWaveformWithTranscription (waveform view) and the in-recorder History button (the "instead of just showing the record button" scenario from the issue).

No types/index.d.ts or provider changes are needed; the data was already on the wire.

### DIFF
--- a/webviews/codex-webviews/src/CodexCellEditor/AudioWaveformWithTranscription.tsx
+++ b/webviews/codex-webviews/src/CodexCellEditor/AudioWaveformWithTranscription.tsx
@@ -256,22 +256,22 @@ const AudioWaveformWithTranscription: React.FC<AudioWaveformWithTranscriptionPro
                     <History className="h-3 w-3" />
                     <span className="ml-1">History</span>
                     {typeof historyCount === "number" && historyCount > 0 && (
-                        <span
-                            className="ml-2 inline-flex items-center justify-center rounded-full"
+                        <Badge
+                            variant="secondary"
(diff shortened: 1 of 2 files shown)


LeviXIII (commented):
Working well. Just a couple of adjustments:

1. The ml-2 should be ml-1 in relation to "History" and the number.
2. This is nit-picky but enough to bother me for a change: when you are in a large screen display, the audioCount/historyCount is centred fine. The moment you shrink the screen down to a smaller viewport, it goes off-center. One thing you can look at doing is actually just using a badge component for this instead of a <span>. The other thing you may need to experiment with is using em instead of rem.

Image

LeviXIII (approved):

PR #1000, opened 2026-05-28
feat(audio): warn and disable recording when no input device detected…

### LINKED ISSUE #980: No audio detected message when input device is not detected
If the user has no input device (microphone) detected, then we should have a warning message that should probably appear when trying to record in the Audio Recording tab of the text editor somewhere. We should also disable recording until a device is detected.

Image

Luke-Bilhorn:
… (#980)

Add useAudioInputDevices hook that enumerates audioinput devices and listens for `devicechange` events so plugging/unplugging a mic updates the UI live. When no device is present, the recorder button is disabled, its tooltip reads "No microphone detected", and an inline warning row appears below the button.

Includes a `window.__forceNoAudioInput` debug flag for QA on machines with a working mic.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2305,6 +2305,7 @@
             "resolved": "https://registry.npmjs.org/@aws-sdk/client-sso-oidc/-/client-sso-oidc-3.609.0.tgz",
             "integrity": "sha512-0bNPAyPdkWkS9EGB2A9BZDkBNrnVCBzk5lYRezoT4K3/gi9w1DTYH5tuRdwaTZdxW19U1mq7CV0YJJARKO1L9Q==",
             "license": "Apache-2.0",
+            "peer": true,
             "dependencies": {
                 "@aws-crypto/sha256-browser": "5.2.0",
                 "@aws-crypto/sha256-js": "5.2.0",
@@ -2358,6 +2359,7 @@
             "resolved": "https://registry.npmjs.org/@aws-sdk/client-sts/-/client-sts-3.609.0.tgz",
             "integrity": "sha512-A0B3sDKFoFlGo8RYRjDBWHXpbgirer2bZBkCIzhSPHc1vOFHt/m2NcUoE2xnBKXJFrptL1xDkvo1P+XYp/BfcQ==",
             "license": "Apache-2.0",
+            "peer": true,
             "dependencies": {
                 "@aws-crypto/sha256-browser": "5.2.0",
                 "@aws-crypto/sha256-js": "5.2.0",
@@ -4982,7 +4984,6 @@
             "integrity": "sha512-T1NCJqT/j9+cn8fvkt7jtwbLBfLC/1y1c7NtCeXFRgzGTsafi68MRv8yzkYSapBnFA6L3U2VSc02ciDzoAJhJg==",
             "dev": true,
             "license": "MIT",
-            "peer": true,
             "engines": {
                 "node": ">=6.9.0"
             }
@@ -5025,7 +5026,6 @@
             "integrity": "sha512-BR7VvDCVHO+q2xBEWskxS6DJE1qRnb7DxzUrogb71CWoSficBxYsiAGd+Kl0mmq/MprG9yArRkyrQxTO6XjMzA==",
             "dev": true,
             "license": "ISC",
-            "peer": true,
             "bin": {
                 "semver": "bin/semver.js"
             }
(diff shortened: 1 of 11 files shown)


Luke-Bilhorn:
image

Luke-Bilhorn:
Claude's instructions for how to test with a working mic (or just ask agent to use forceNoAudioInput lol)

## Testing the no-microphone UI with a working mic

1. Pull this branch
2. Open any cell → switch to the **Audio** tab.
3. Open the Command Palette (`Cmd+Shift+P`) → run **`Developer: Open Webview Developer Tools`**.
4. In the devtools Console that opens, paste:

   ```js
   window.__forceNoAudioInput = true;
   navigator.mediaDevices.dispatchEvent(new Event("devicechange"));
   ```

5. The mic button should immediately turn solid grey with a slashed-mic icon, and a yellow alert appears below: *"No microphone detected. Connect an input device to record."*

To restore the normal state:

```js
window.__forceNoAudioInput = false;
navigator.mediaDevices.dispatchEvent(new Event("devicechange"));
```

The flag is intentionally left in production code (documented in `hooks/useAudioInputDevices.ts`) and does nothing unless explicitly set, so it's safe to ship. Useful for ongoing QA on machines with working mics.

LeviXIII (commented):
Doesn't seem to work for me. I tried removing permissions (which would be an ideal case for this warning to come up for as well) and disabling using the debug commands. I am working on a mac. 

Also, please check the padding difference from the top to the bottom of the warning section. Also the margin between the icon and the text in the warning message as it is also not vertically aligned.

Luke-Bilhorn:
image

Luke-Bilhorn:
image

Luke-Bilhorn:
Matt, I have fixed all of the things we discussed:
- warning spacing
- mic countdown issues
- separate messages for no mic connected vs no mic permission

I have tested it and found that we are now correctly getting the OS-level permission (see last commit message), but I have no way to verify that we will be correctly getting the OS-level state for a device with no mic. I think it's pretty likely we are because it is the same type of thing, but this is information I can't learn from my own device. Even if we can't test this before Thursday, I think we should still merge it because the case about no permission is way more important.

What Claude says:
Chromium's Permissions API only reports browser-level state, not OS-level permission. On macOS, for example, denying microphone access in System Settings still returns "granted" from navigator.permissions.query — so the hook never flipped to permission-denied without an actual record attempt.

Expose `reportRecorderError(err)` from useAudioInputDevices and call it from startActualRecording's catch block. The hook classifies common error names (NotAllowedError, NotFoundError, etc.) and pins availability to the matching state. A successful getUserMedia clears the override so the UI recovers automatically after the user fixes permissions and retries.

Luke-Bilhorn:
## Testing

The yellow warning has two variants: **"No microphone detected"** (no audio input hardware) and **"Microphone access denied"** (a mic exists but is blocked). Both render the same grey, slashed-mic button.

To force either state without unplugging hardware or revoking permissions, edit **line 41** of `webviews/codex-webviews/src/CodexCellEditor/hooks/useAudioInputDevices.ts`:

```ts
const FORCE_STATE_FOR_REVIEW: MicAvailability | null = null;
```

Replace `null;` with one of:
```ts
"no-device";
"permission-denied";
```

Then rebuild and reload:

```bash
cd webviews/codex-webviews && pnpm run build:CodexCellEditor
```

Hit `Cmd+R` in the Extension Development Host window. Set the constant back to `null` before merging.

Once forced into either state, also verify:
- Clicking the disabled mic button does nothing (no countdown, no `getUserMedia` call)
- If the auto-start-on-cell-open feature is enabled, opening a cell with no mic does NOT start a countdown

LeviXIII (commented):
The button doesn't disable if Codex doesn't have permission set. I understand that Claude says it can do it through Chromium only, but that isn't entirely true. We can have it detect the OS level permission before opening the Audio Recording tab. We may have to cache the result if it results in opening the permission modal every time we click, but it shouldn't. I'll check to see if this works on a Windows machine as well.

Luke-Bilhorn:
The reason it didn't appear disabled before was that we waited for the user to click the record button before checking mic access, so the button stayed enabled until the first failed click. I've added a probe that runs as soon as they open the Audio tab, so if they can't record, the button disables immediately.

Luke-Bilhorn:
And we cache that result per session

LeviXIII (dismissed):
Works well.

LeviXIII (approved):

LeviXIII:
Additional fixes slipped into this PR: Delete button handling in the project card (no longer allows for multiple clicks)

PR #1004, opened 2026-05-29
904 feature request randall eten make options for cluster or individual exporting

### LINKED ISSUE #904: feature request: Randall ETEN:  Make options for cluster or individual exporting
- Randal says "I am trying to share the audio for the first 2 chapters of Matthew with someone external. When I tried, I found the following challenges: 1. When I export in Codex, I can only download entire books. 2. Then, the files are essentially verse by verse, but the file naming doesn't tell me which file is which chapter & verse. 3. Also, because I only have the project on Stream, it looks like the export only exports the pointers and not the actual audio files. "

- Problem: We cannot export and share the actual audio for only Matthew 1–2 because Codex forces whole-book export, outputs unlabeled verse files, and from Stream exports only pointers instead of audio.
- Context: We need to send just the first two chapters of Matthew to an external recipient. The project is only on Stream, and Codex exports are verse-by-verse with non-descriptive filenames and do not include the underlying audio files.
- Interface URL: https://airtable.com/appfEUXIPganSIP89/pagy8X7KblenRWOQD?detail=eyJwYWdlSWQiOiJwYWdlMFZ4ZjlrWGQ4V3EyTiIsInJvd0lkIjoirec0Ot8II8rcZAkjVIiwic2hvd0NvbW1lbnRzIjpmYWxzZSwicXVlcnlPcmlnaW5IaW50Ijp7InR5cGUiOiJwYWdlRWxlbWVudCIsImVsZW1lbnRJZCI6InBlbE1FNjdwUUVYZVowZEFJIiwicXVlcnlDb250YWluZXJJZCI6InBlbFBNNERmtE43aUF6T3Y1eCJ9fQ==

We will split this into 3 tickets that need to be completed in order.

Fikitti:
added ability for bible files to export audio in "milestones" (chapters). there will be a third step to select what milestones to include for audio export (it will not mess up the text export, that will work normally for the whole file as it always does).


LeviXIII (commented):
Functionality is good. There is one issue and one text suggestion that I found though:

1. For older projects, I don't get the step for milestone exporting. We need to have a way for older projects to see this step. It's hit or miss as both projects didn't have an importerType, but one was able to see the step while the other wasn't.

2. For the Select Milestones description, we don't need to include "All milestones are selected by default" as we can see all the checkboxes checked off and the "Select All" checkbox checked. You can remove that sentence.

Screenshot 2026-06-01 at 7 33 38 AM

LeviXIII (commented):
One new issue:
- If I have 2 books that are exporting audio (Gen and Mat) and then I unselect all the books from Mat. and just take one chapter from Gen, Mat still exports the first milestone available.

Image

Image

Image

LeviXIII (commented):
2 John doesn't seem to be give me an option to download by milestone while 2 Peter does. In fact, you may want to take a look at books with just one chapter as they seem to be acting up (eg. Jude gives me two chapter 1s)

Image

Image

LeviXIII (approved):

PR #1024, opened 2026-06-03
fix(editor): restore unmerge option + in-editor merge confirmation (#691)

### LINKED ISSUE #691: Bug: merged content no longer shows the option to unmerge
Image

Unmerging cells is not rendering as an option. I have merged some of the cells when in "Source Editing Mode" (runs the Source Editing Mode command to activate this feature).
After merging content I could see an option to unmerge aka revert the action. Clicking it will add a new edit to the cell and mark it as unmerged. 
Note: it only marks the cell as merged: false. It does not revert the change to the cell above. This is fine. There is a bunch of complications at play so reverting the full action is a multi step process that the user needs to figure out. 

We need to render the revert merge option again. The ui is still in the code. But for some reason it doesn't display.

sampjvv:
## What & why

Fixes #691 — _"merged content no longer shows the option to unmerge"_ — plus a UX follow-up on the merge confirmation.

### 1. Unmerge button no longer showed (the reported bug)
In Source Editing (correction) mode the provider intentionally keeps merged cells in the translation units so they can be unmerged, and `CellContentDisplay` renders a **Cancel merge** button on them. But `CellList` was filtering merged cells out in **exactly that mode** (`filteredTranslationUnits`), so the merged cell never reached `CellContentDisplay` and the button could never render.

The filter contradicted the rest of `CellList`, which already styles merged cells (greyed at `opacity: 0.5`, `❌` line number). Removing it restores the intended behavior; it's safe because the provider already strips merged cells in every other mode.

### 2. Merge confirmation moved into the editor
Confirming a merge used a native `vscode.window.showWarningMessage`, which pops up in the **bottom-right corner of the screen** — far from the cell — and wrongly said the action _"cannot be undone"_ (merges are reversible via unmerge).

It's now a small **in-editor modal** (the same `Dialog` pattern as the existing auth/offline modals) with reversible wording and **Cancel** / **Merge** buttons. `confirmCellMerge` is only posted once the user confirms; the handler no longer shows the native dialog (its child-cell guard is preserved), and the now-unused `message` payload field is removed from the type.

> Note: per the ticket, unmerge intentionally only re-exposes the cell (sets `merged: false`); it does **not** reverse the content concatenation. That behavior is unchanged here.

## Changes
- `webviews/.../CodexCellEditor/CellList.tsx` — remove the merged-cell filter.
- `webviews/.../CodexCellEditor/CellContentDisplay.tsx` — in-editor merge-confirmation modal.
- `src/providers/codexCellEditorProvider/codexCellEditorMessagehandling.ts` — drop the native confirmation from `confirmCellMerge`.
- `types/index.d.ts` — remove the unused `message` field from `confirmCellMerge`.
- Tests: `CellList.mergedCellUnmerge.test.tsx`, `CellContentDisplay.mergeConfirm.test.tsx`.

### Unmerge option restored
- [x] After merging cells in **Source Editing Mode** (correction editor mode for source text), the merged cell still renders (greyed-out) and shows the **unmerge / revert merge** button
- [x] Clicking unmerge adds a new edit to the cell and marks it `merged: false`
- [x] Unmerging only marks the cell as unmerged — it does **not** attempt to revert the content of the cell above (expected behavior)
- [x] Merged cells are **not** hidden in source + correction editor mode (the old `cellData?.merged` filter no longer strips them)
- [x] In all other (non-correction) modes, merged cells remain absent from the list because the provider strips them upstream

### In-editor merge confirmation modal
- [x] Clicking **Merge** opens an in-editor confirmation dialog near the cell (no longer the native VS Code bottom-right popup)
- [x] The dialog title/description explain the merge and that it can be reversed via unmerge
- [x] Clicking **Merge** in the dialog posts `confirmCellMerge` and closes the modal
- [x] Clicking **Cancel** (or dismissing via overlay/Esc) closes the modal and posts **nothing**
- [x] No merge occurs until the in-editor dialog is confirmed

### Message contract / provider (regression)
- [x] `confirmCellMerge` payload no longer includes a `message` field (`types/index.d.ts` and webview agree)
- [x] `confirmCellMerge` handler proceeds directly to merge without showing a second native `showWarningMessage` confirmation
- [x] Cells with child cells still follow the existing pre-merge handling path (not bypassed)
- [x] For **source files**, merging also merges matching cells in the target file (`mergeMatchingCellsInTargetFile`)
- [x] For **non-source files**, the target-file merge step is skipped
- [x] Merge errors are caught and logged without crashing the editor

### DIFF
--- a/src/providers/codexCellEditorProvider/codexCellEditorMessagehandling.ts
+++ b/src/providers/codexCellEditorProvider/codexCellEditorMessagehandling.ts
@@ -3121,7 +3121,7 @@ const messageHandlers: Record<string, (ctx: MessageHandlerContext) => Promise<vo
 
     confirmCellMerge: async ({ event, document, webviewPanel, provider }) => {
         const typedEvent = event as Extract<EditorPostMessages, { command: "confirmCellMerge"; }>;
-        const { currentCellId, previousCellId, currentContent, previousContent, message } = typedEvent.content;
+        const { currentCellId, previousCellId, currentContent, previousContent } = typedEvent.content;
 
         debug("confirmCellMerge message received for cells:", { currentCellId, previousCellId });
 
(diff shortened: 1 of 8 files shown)


LeviXIII (commented):
When unmerging, in the json for the file, the author is listed as "anonymous". It should have the name of the user that unmerged it like it does for the one who did merge it:

{
            "editMap": [
              "metadata",
              "data",
              "merged"
            ],
            "value": true,
            "timestamp": 1782495713614,
            "type": "user-edit",
            **"author": "matthew_test",**
            "validatedBy": []
          },
          {
            "editMap": [
              "metadata",
              "data",
              "merged"
            ],
            "value": false,
            "timestamp": 1782495774614,
            "type": "user-edit",
            **"author": "anonymous",**
            "validatedBy": []
          }

sampjvv:
Fixed in caf07982: unmerge now records the acting user instead of hardcoded `anonymous` (matches the merge edit). Verified live — both target and source unmerge edits show the real user — and added a test.

LeviXIII (approved):

PR #1025, opened 2026-06-03
fix(navigation): open the correct chapter/verse when deep-linking into a not-yet-open book (#996)

### LINKED ISSUE #996: When clicking on cells without audio proper chapter doesn't open
It will open the correct book, but the chapter and verse will not be there.

https://github.com/user-attachments/assets/d88cdc6a-d789-43ca-826e-3719c12e9f7a

sampjvv:
## What & why

Fixes #996 — _"When clicking on cells without audio, the proper chapter doesn't open."_

In the export wizard's Step 1 audio-stats popover (which lists a book's **text-only** cells), clicking a cell is supposed to open the source+target editors and navigate to that cell's chapter/verse. **When the book was not already open**, it instead opened at **chapter 1** (e.g. clicking 11:20 → chapter 1). When the book was already open, it worked — which made it intermittent.

## Root cause (cold-open race)

`openCodexDocumentWithSourcePair` opened the **source** pane and `await`ed `waitForWebviewReady(source)`, then opened the **target `.codex`** pane **without waiting** for it. Callers (the export popover via `jumpToCellInNotebook`) publish the navigation jump — the `cellToJumpTo` workspace-state key — immediately after this helper returns.

The per-editor `cellToJumpTo` listener is registered while the editor *resolves*, and `waitForWebviewReady` only returns once a pane has signalled ready (i.e. after it resolved + registered its listener). With the target pane not waited for, a **cold-opened** editor could miss the jump and fall back to its default initial page (chapter 1). The general jump path was hardened for the search view back in March, but this helper's missing target-pane wait left the export deep-link (added shortly before this was reported) exposed.

## Fix

Resolve the editor provider once and `await waitForWebviewReady` for **both** panes (source *and* target) before returning — in the normal path, the source-failed fallback, and the no-workspace path. Cold-open now behaves like the working warm-open.

The helper is shared (navigation sidebar, NewSourceUploader "Start translating"), so they get the same small target-readiness wait and the same cold-open race protection; no other behavior change.

## Changes
- `src/utils/openCodexDocumentWithSourcePair.ts` — wait for the target pane's webview ready (plus a single provider lookup + small `waitForPaneReady` helper).

## Testing
- `npm run compile` clean; typecheck clean for the changed file.
- Manual (the confirmed repro): reload the Extension Dev Host → **close all editor tabs** → export wizard → click a chapter-2+ text-only cell (e.g. 11:20) → opens both panes at the correct chapter/verse, not chapter 1.
- Regression: warm open (book already open) still correct; navigation-sidebar verse open and "Start translating" still open the pair correctly.


LeviXIII (commented):
Unfortunately, this fix did not work. For example, when I attempt to open Exodus 11:1 and the Exo.codex file isn't already opened, it goes to Exodus 1:1. It's still not getting to the correct chapter.

LeviXIII (approved):

PR #1032, opened 2026-06-05
Adding a delete confirmation for urls and video files

TimRl:
## 1031-Adding a delete confirmation for urls and video files
#1031 

Project: 1032-delete-video-safely

## Summary

Removing or replacing a chapter video previously took effect with only a single, easily-dismissed native confirmation — risky because removing a local video permanently deletes the file from the project on save. This PR adds a robust "type-to-confirm" safety gate in the Edit Metadata modal: to remove a video, the user must type or paste the exact file name (local) or URL (remote) before the removal is enabled. It also removes the now-redundant second confirmation that fired afterward, so the user confirms exactly once.

## Changes

- Clicking **Clear** on a video in the Edit Metadata modal no longer removes it immediately; it opens a destructive confirmation panel.
- The user must type or paste the exact **file name** (local video) or **URL** (remote video) to enable the **Remove video** button; matching is exact (whitespace-trimmed) so a pasted value with stray spaces still works.
- The expected value is shown in a selectable/copyable field, with a "doesn't match yet" hint while typing a non-matching value.
- Confirming only empties the draft; the actual file deletion + metadata change still happens on **Save Changes** (and Cancel/X discards everything) — the existing deferred-save model is preserved.
- Keyboard support: `Enter` confirms when matched; `Escape` cancels just the confirmation without closing the modal; the input auto-focuses.
- The confirmation state resets when the modal opens, when the video reference changes (e.g. picking a new file), and on close.
- Removed the redundant host-side "Replace/Remove… will be deleted" prompt for modal-driven changes (a new `skipVideoConfirm` flag), since the modal now performs a stronger confirmation. The host still deletes the old file, and keeps the prompt as a fallback for any non-modal caller.

## Testing Checklist

- [x] **Local file – remove:** Clear a local video → confirmation panel appears showing the file name; **Remove video** is disabled until the typed/pasted name matches exactly.
- [x] Typing a wrong/partial name keeps **Remove video** disabled and shows "The text doesn't match yet."
- [x] Pasting the file name (with stray leading/trailing spaces) enables removal.
- [x] **URL – remove:** Clear a URL video → must type/paste the full URL to enable removal.
- [x] Confirm removal → field empties, but nothing is deleted until **Save Changes**.
- [x] **Save Changes** after a confirmed local removal → the local file is actually deleted from `files/` and `pointers/`; "Show Video" toggle disappears.
- [x] **Cancel/X** after confirming removal (before Save) → video intact; local file **not** deleted; player still works.
- [x] No redundant second native "Remove?/Replace?" dialog appears on Save after the in-modal confirmation.
- [x] **Replace flow:** Clear (type-to-confirm) → pick a new file → no extra "Replace?" prompt; the old local file is deleted and the new one is saved.
- [x] **Esc** inside the confirmation input cancels only the confirmation and does **not** close the modal.
- [x] **Enter** in the input removes the video only when the text matches.
- [x] Opening the modal / switching reference resets any in-progress confirmation (no stale panel).
- [x] Regression: editing non-video fields (font size / direction / corpus marker) and saving still works with no video prompts.

### DIFF
--- a/src/providers/codexCellEditorProvider/codexCellEditorMessagehandling.ts
+++ b/src/providers/codexCellEditorProvider/codexCellEditorMessagehandling.ts
@@ -1783,7 +1783,12 @@ const messageHandlers: Record<string, (ctx: MessageHandlerContext) => Promise<vo
             const oldKind = classifyVideo(oldVideoUrl);
             const newKind = classifyVideo(newVideoUrl);
             if (oldKind === "local") {
-                const proceed = await confirmVideoReplacement(oldKind, newKind);
+                // The metadata modal already runs a robust type-to-confirm step
+                // before any video removal/replace, so it sets skipVideoConfirm to
+                // avoid a redundant second prompt. Other callers still confirm.
+                const proceed = typedEvent.skipVideoConfirm
+                    ? true
+                    : await confirmVideoReplacement(oldKind, newKind);
                 if (!proceed) {
                     // Revert the webview's optimistic edit by re-sending current
                     // metadata, and restore the player's URL to the unchanged video.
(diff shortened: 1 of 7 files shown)


LeviXIII (commented):
A couple of issues:

1. Selecting "Esc" when in the input box for the removal confirmation DOES close the entire modal. It should just close the removal portion.

2. Trying to type in the input just causes the first letter to be saved to the input

Image

LeviXIII (approved):

PR #1041, opened 2026-06-09
InDesign import export update

### LINKED ISSUE #716: Make a new importer/exporter for Reach 4 Life (Biblica) files
Reach 4 Life (also IDML) files importer and exporter should function the same way as the Biblica Study Notes one is working fine tuned for the new files.

Fikitti:
Developed a new way for regular InDesign importer exporter to maintain the correct strucutre and surgically replace the correct textual content in between all the styles and tags. Work more reliable than the older version. Now it just saves you could say "end-of-content" or "style-block" with no further explanation and preserving them as sort of "text splitters". When the exporter will be replacing the content and it will hit this splitter, it will then skip all the styling in the original document until it finds new text it starts replacing until it hits new splitter and so on.

## PR Title
Format:
[Issue Number]-[Issue Name]

## Summary

Closes #716 
This might not completely close the issue, but it's a great step towards that, I will still have to run more tests with files they will provide me later.

## Testing Checklist
- [x] import regular indesign file
- [x] translate cells also with html enforcement necessary
- [x] export the file and see if it was exported correctly with preserved locations, styles etc.

### DIFF
--- a/debug.log
+++ b/debug.log
@@ -1,0 +1,2 @@
+[0617/150301.373:ERROR:third_party\crashpad\crashpad\util\win\registration_protocol_win.cc:108] CreateFile: The system cannot find the file specified. (0x2)
+[0617/150302.820:ERROR:third_party\crashpad\crashpad\util\win\registration_protocol_win.cc:108] CreateFile: The system cannot find the file specified. (0x2)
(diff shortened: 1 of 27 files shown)


LeviXIII (approved):
Works well!

PR #1061, opened 2026-06-20
fix: stream audio for Consolidate by Character export in stream-only …

### LINKED ISSUE #1060: Bug: Audio Exports for Consolidate by Character only work in auto-download mode
## Description
There's no way to export audio when using the Consolidate by Character option in stream only. We recently changed the audio exporter to be able to download while in this mode. The Consolidate by Character exporter needs to adopt the same method so people don't have to be in auto-download mode.

## Steps to Reproduce
1. Set some labels on a .vtt file and record some audio for those cells so that it can split by label names when exporting
2. Go to export and choose the .vtt file as well as the Consolidate by Character option
3. Try to export and you should get the 0 files written message.

## Expected Behavior
The Consolidate by Character option should work just like the Include audio and Include Audio with Timestamp options when it comes to being in stream only mode. So the audio should only download when exporting and have the option to do it per milestone.

## Actual Behavior
The Consolidate by Character option doesn't allow any audio to be downloaded while in stream only mode.

## Acceptance Criteria
- [x] **Stream-only export works.** Exporting with the **Consolidate by Character** option succeeds when the project is in `stream-only` mode — audio is downloaded on demand during export rather than requiring `auto-download` / locally-present files. No more "0 files written" message.
- [x] **Parity with other audio modes.** Consolidate by Character behaves the same as **Include Audio** and **Include Audio with Timestamps** with respect to media strategy — it downloads audio at export time via the Frontier API (LFS), not before.
- [x] **Per-milestone support.** Audio downloads/exports respect the selected milestones (`selectedMilestonesByFile`); only audio for selected milestones is downloaded and written.
- [x] **All media strategies handled.** Works under `stream-only`, `stream-and-save`, and the existing locally-available (download) modes.
- [x] **Character split preserved.** Output still produces one file per character label, all clips aligned to 0:00 for DAW alignment — the streaming change does not alter the consolidation/mixing behavior.
- [x] **Graceful failure.** If Frontier is unavailable or a version gate fails, the export fails with a clear error message (consistent with `exportAudioAttachments`), not a silent skip or "0 files written."

### Core Scenario (Repro Path)
- [x] Set character labels on a `.vtt` file and record audio for those cells so it can split by label name.
- [x] Put the project in **stream-only** mode (audio stored as LFS pointers, not local).
- [x] Go to export, select the `.vtt` file and the **Consolidate by Character** option.
- [x] Export → confirm files are written (no "0 files written" message).
- [x] Confirm one output file per character label, each starting at 0:00.

### Media Strategy Variants
- [x] **stream-only** → audio downloads during export and writes correctly.
- [x] **stream-and-save** → audio downloads, exports, and is saved locally as expected.
- [x] **Local / auto-download (files already present)** → still works (regression — no double-download).
- [x] Mix of locally-present and pointer-only audio in one export → all resolve correctly.

### Per-Milestone Behavior
- [x] Select a subset of milestones → only audio for selected milestones is downloaded and exported.
- [x] Select all milestones → all character audio exported.
- [x] Deselect a file entirely (empty milestone array) → that file is skipped, no downloads for it.
- [x] Character clips spanning multiple milestones are grouped/exported correctly per the selection.

### LFS / Download Resolution
- [x] Audio resolved from `/.project/attachments/files/` when present.
- [x] LFS pointer in `files/` triggers download via Frontier (`downloadLfsWithRetry`).
- [x] Fallback to `/.project/attachments/pointers/` when `files/` entry is absent.
- [x] In-memory cache avoids re-downloading the same clip used by multiple character tracks.
- [x] Concurrency pool downloads multiple clips in parallel without corruption or race conditions.

### Output Correctness
- [x] Each character file mixes the correct clips for that label (FFmpeg render with downloaded bytes).
- [x] All clips start at 0:00 (DAW alignment preserved).
- [x] Consolidated audio format option (wav / flac / opus) honored.
- [x] Output folder/milestone structure matches the other audio export modes.

### Error / Edge Handling
- [x] Version gate failure → graceful, descriptive error (consistent with `exportAudioAttachments`).
- [x] Download retry/backoff works on transient network failure.
- [x] Cancelling the export mid-download aborts in-flight downloads (no dangling activity).
- [x] Cell with a label but no recorded audio → handled gracefully (skipped, no crash).
- [x] Character label with no clips in the selected milestones → no empty/broken file produced.

### Regression
- [x] **Include Audio** export still works in stream-only and local modes.
- [x] **Include Audio with Timestamps** still works (WAV conversion + timecode metadata intact).
- [x] Non-audio exports (text/`.vtt` only) unaffected.
- [x] No console errors or unhandled promise rejections during a successful consolidate-by-character stream-only export.

Luke-Bilhorn:
## PR Title
1060-Audio Exports for Consolidate by Character only works in auto-download mode

## Summary

Closes #1060

Makes Consolidate-by-Character exporting work with all stream/download modes.

Claude: "Consolidate by Character now resolves audio via the same LFS streaming path as Include Audio, so export works in stream-only/stream-and-save (not just auto-download)."

## Changes

- Extracted shared `setupAudioStreaming` + exported `resolveAudioBytes` from `audioExporter.ts`
- Character export downloads LFS clips to temp files for FFmpeg, with retry/cache/concurrency
- Honors `selectedMilestonesByFile`; supports cancellation
- Clear error when streaming setup or all downloads fail (no silent "0 files written")

### Consolidate by Character
- [x] Works in stream-and-save with LFS pointers (no "0 files written")
- [x] Works in auto-download with local files
- [x] One `.wav` per character label; clips DAW-align from 0:00
- [x] Per-milestone selection respected

### Regression
- [x] Include Audio still works in stream-and-save
- [x] Include Audio with Timestamps still works (`.wav` output)

## Screenshots

N/A


LeviXIII (commented):
Instead of having the toast with the progress when preparing the content, it should follow the same pattern of having the Download box show what is happening. The toast stops after the content is prepared, but I don't know how the other processes are moving. If I cancel, it does stop everything, but it should show me at which point I am cancelling.

Basically, make it work like it does for the other audio formats.

Image

LeviXIII (commented):
This also is not working: Download retry/backoff works on transient network failure.

Need to go though the test checklist on the main ticket to make sure those points are still working.

Luke-Bilhorn:
I think all of this is fixed... and I did an odious amount of testing to make sure. 

Haven't yet tested multi-milestone consolidation (didn't have a file for it, but I could go figure out how to make a vtt for this) or when frontier unavailable (It was still working somehow when I was signed out). But other than these it seems pretty solid.

LeviXIII (approved):
Works well now. If a character is not in the milestone section that is chosen and doesn't have any other speaking parts in the chosen milestone, it will not write out their file.

PR #1066, opened 2026-06-23
1065-burmese-bib-project-duplicate-cells-issue

### LINKED ISSUE #1065: Burmese bib project: duplicate cells issue
- Problem: A recurring “Duplicate cells found” error page in the editor blocks the Burmese team from accessing verses, creating a loop that halts translation progress.
- Context: All Burmese team members see this overlay in the Burmese project, and it reappears immediately after navigating back to the chapter. Unclear checkmark/X prompts prevent resolving the issue, leaving them stuck.

## Description
Some books show a "Duplicate cells found" screen instead of the chapter text. This happens because a few section-heading cells got copied — the file ends up with two identical cells that share the same ID.

The cause is the migration that gives every cell a unique ID. It creates each ID by hashing the old ID, so the same input always produces the same output. If a heading already exists twice in the file (once with an old-style ID, once already
converted), running the migration again turns the old one into the exact same ID as the converted one — so now there are two identical cells.

This shows up mostly in Genesis, Exodus, Leviticus, Numbers, and Deuteronomy, because those books have section headings that went through this conversion.

## Steps to Reproduce
1. Open a freshly cloned/synced copy of the project (so migrations run again).
2. Open an affected book, e.g. `DEU.codex`.
3. Click forward in the chapter navigation to the first chapter with a duplicated heading (Deuteronomy 3).
4. The "Duplicate cells found" screen appears instead of the chapter.

## Expected Behavior
- Running the ID migration never creates two cells with the same ID.
- If a heading already exists in both old and new ID form, the migration keeps just one copy.
- Every chapter opens normally.

## Actual Behavior
- The migration gives the old-style heading the same ID as the already-converted one, creating an exact duplicate.
- Nothing cleans this up afterward, so the duplicate gets saved and synced.
- The editor blocks the chapter with "Duplicate cells found".

## Acceptance Criteria
- [ ] The ID migration checks for existing IDs and never creates a duplicate; if the new ID already exists, it merges into that cell instead.
- [ ] Running the migration twice on the same file produces no duplicates.
- [ ] Merging two copies keeps all edit history (no data loss).
- [ ] Already-affected books (GEN/EXO/LEV/NUM/DEU) are repaired.
- [ ] Opening any chapter no longer shows "Duplicate cells found".

## Test Checklist
- [ ] A file with a heading in both old and new ID form ends up with one cell after migration.
- [x] Running the migration twice changes nothing the second time.
- [ ] The surviving merged cell keeps its edit history and validations.
- [ ] A file that's already fully converted is left untouched.
- [ ] After repair, navigating every chapter of GEN/EXO/LEV/NUM/DEU shows no duplicate screen.

Image

- Interface URL: https://airtable.com/appfEUXIPganSIP89/pagyewiWtbGAM67M4/recILHueXyuaUPMHf?home=pagy8X7KblenRWOQD

LeviXIII:
## Summary
Add collision guard and merge logic for duplicate cell IDs in migration process

Closes #1065

## Changes
- This update introduces a new function to merge cells with duplicate IDs during the migration of cell IDs to UUIDs.
- It ensures that cells with the same logical ID are combined, preserving order and edit history, to prevent duplicates from being emitted during the migration process.
- Additionally, a one-time repair sweep is added to clean up any existing duplicates across notebooks, enhancing data integrity.

## Test Checklist
- [x] A file with a heading in both old and new ID form ends up with one cell after migration.
- [x] Running the migration twice changes nothing the second time.
- [x] The surviving merged cell keeps its edit history and validations.
- [x] A file that's already fully converted is left untouched.
- [x] After repair, navigating every chapter of GEN/EXO/LEV/NUM/DEU shows no duplicate screen.


sampjvv:
## QA summary

**Code review**: the collision guard sits inside `migrateCellIdsToUuidForFile` so the migration can no longer emit same-id duplicates; the one-time startup sweep heals already-affected projects; the merge combiner reuses the sync resolver logic (latest-edit-wins on value, edit histories unioned, attachments/validations preserved). New test suite passes 6/6 locally.

**Live testing** (forked Tripuri test project, real data): planted the three failure shapes and confirmed the startup path auto-repaired all of them on open, with zero collateral changes to the other 13 notebook files:
1. **Byte-identical same-id twins** → merged to one cell, file byte-identical to pre-duplication state. No "Duplicate cells found" screen.
2. **Divergent same-id twins** (copies edited differently) → survivor shows the newer text, both edit histories preserved in `edits`.
3. **Legacy paratext id + its already-hashed UUID twin** (the exact #1065 mechanism) → migration collapsed them into one cell carrying both edit histories, correct hashed UUID, no duplicate emitted.

Two observations, neither blocking:
- "Codex: Merge Duplicate Cells" also normalizes **all** legacy ids to UUIDs per file (via `migrateCellIdsToUuidForFile`), which is more than the command name suggests — fine, just worth knowing it can produce large diffs on un-migrated projects.
- A teammate on an **old build** can still create + sync in a fresh duplicate after a workspace's one-time sweep flag is set; nothing automatic cleans that up (sweep is startup-only by design) — the manual command is the escape hatch, so affected teams should update together.

Also noteworthy: in the test forks, `.vscode/settings.json` is committed **with** `cellIdsToUuidMigrationCompleted: true`, so fresh clones of those projects don't re-run the UUID migration at all — projects where that flag isn't committed are the ones that hit #1065.

Approving and merging.

sampjvv (approved):
Reviewed the code and verified all three repair paths against a real test project — solid fix.

PR #1068, opened 2026-06-23
Fix #1055: include verse ranges in audio export filenames

### LINKED ISSUE #1055: Bug: Export files don't account for verse ranges in file name
## Description

When exporting an audio file, our new format is book, chapter, verse, line. However, when there is a verse range, it doesn't get listed in the filename.

## Steps to Reproduce

1. Export a file with audio and make sure there is a verse range in it

## Expected Behavior

Verse range should be a part of the file name like V1-2 (1PE_C3_V1-2_L51)

## Actual Behavior

It just prints as 1PE_C3_L51

## Acceptance Criteria
- [x] Exporting a file with a verse range contains the actual verse range
- [x] The audio in the verse range will play correctly
- [x] Single verses are formatted correctly
- [x] There won't be any duplicate verses due to the change

### Filename format
- [x] Single verse exports as `BOOK_C{chapter}_V{verse}_L{line}` (e.g. `1PE_C3_V1_L51`)
- [x] Verse range exports with both endpoints (e.g. `1PE_C3_V1-V2_L51` per the expected format)
- [x] Confirm the exact range separator/format is consistent everywhere it's generated (decide `V1-V2` vs `V1-2` and verify one is used)
- [x] Book, chapter, verse, line order is preserved with the range inserted in the verse position

### Range scenarios
- [x] Two-verse range (e.g. V1-V2)
- [x] Multi-verse range spanning 3+ verses (e.g. V1-V5)
- [x] Range spanning a chapter boundary (if supported) — verify naming is sane or handled
- [x] Range with non-sequential / padded verse numbers (e.g. V001-V010) formats correctly
- [x] Mixed export containing both single verses and verse ranges in the same file/batch

### Regression
- [x] Single-verse (non-range) filenames are unchanged by the fix
- [x] Books with numeric prefixes (1PE, 2CO, etc.) still format correctly
- [x] Line number (`L{n}`) still appended correctly after the verse/range
- [x] No double underscores, missing segments, or malformed names

Image
Image

sampjvv:
## Summary

Fixes #1055. Audio exports dropped the verse **range** from filenames: a cell spanning verses 1–2 exported as `1PE_C3_L51` instead of `1PE_C3_V1-2_L51`.

**Root cause:** in `src/exportHandler/audioExporter.ts`, `parseCellIdToBookChapterVerse()` read the ref (e.g. `"1PE 3:1-2"`), split on `:` to get `"1-2"`, then did `Number("1-2")` → `NaN`, which was discarded — leaving only the chapter.

## Changes

- Reuse the existing, range-aware `parseVerseRef()` from `verseRefUtils` (it already returns `verseStart`/`verseEnd` and strips legacy cell-id suffixes). A `fallbackParseRef()` preserves today's behavior for chapter-only / malformed ids.
- `parseCellIdToBookChapterVerse()` now returns `verseEnd`; `formatChapterVerseSuffix()` emits `C3_V1-2` for ranges and `C3_V1` for single verses (a degenerate `1-1` collapses to `V1`).
- **Also covers UI-merged cells.** No merge path writes the range into `globalReferences` — it only sets the kept cell's `cellLabel` to `1-2`. New `rangeEndFromCellLabel()`: when only a single verse is resolved from the ref, derive the range end from a clean `"{start}-{end}"` cellLabel whose start matches that verse. So **both** imported range cells (`globalReferences` = `"1PE 3:1-2"`) **and** merged cells (single ref + `"1-2"` label) export as `1PE_C3_V1-2_L51`. `isExportableCell` already skips `merged`/`deleted` cells, so only the kept cell exports (no duplicate verses).
- Exported the helpers and added unit tests in `audioExporter.test.ts`.

**Filename format:** `V1-2` (start-end, single `V`) per the decision on the issue.

## Tests

The isolated `audioExporter` suite passes **31/31** (`CODEX_TEST_GLOB=audioExporter`), incl. 18 #1055 cases: single verses (regression), 2-verse and 3+ ranges, numeric book prefixes (`2CO`), padded-number normalization, cellId fallback, chapter-only/malformed refs, and the merged-cell label cases (single ref + range label, multi-verse label, start-mismatch guard, plain single label, globalReferences-takes-precedence). `tsc --noEmit` and ESLint clean.

## Known limitations (non-blocking)

- **Cross-chapter ranges** (e.g. `GEN 1:31-2:1`) aren't representable in a single cell's ref and degrade to chapter-only (`C1`).
- **Padded verse numbers** normalize (leading zeros stripped, e.g. `V001-010` → `V1-10`).

> Note: the automated tests cover filename generation; the manual checklist item "audio in the range plays correctly" still needs a human spot-check in the extension host.


LeviXIII (approved):
Good change. No problems with multi-verse range filenames.

PR #1072, opened 2026-06-26
Fix cell list audio status

TimRl:
## Summary

Fixes audio download behavior so the app respects the "auto-download" setting,
clearly shows when audio is downloading, and updates the audio icons instantly
instead of only after a refresh.

## Changes

- **Showing the proper state of audio files in the cell list** Whether the file is available locally or available to download, it'll now be reflected properly (in Windows it wasn't always exact due to the OS file path structure.

- **Icons update the moment a file arrives.** When audio finishes downloading in the
  background, the cell's icon switches to playable right away, without a refresh.

- **Stop unwanted downloads when browsing history.** Picking a different recording
  in the history no longer downloads it when auto-download is turned off.

- **Show a download in progress when picking from history.** With auto-download on,
  choosing a recording that isn't saved yet shows a downloading state in both the
  history list and the cell editor.

- **Clearer wording.** The history list now says "Downloading…" instead of "Loading…".

- **The cell list reflects downloads you start in the editor.** Click "Download" in a
  cell, then go back to the list, and that cell now shows it's downloading.

## Testing Checklist

- [x] With auto-download off, switching between recordings in the history does not
      download them.
- [x] With auto-download on, picking an unsaved recording shows a downloading state
      in the history list.
- [x] With auto-download on, the cell editor shows "Downloading audio…" and then
      plays the recording.
- [x] The history list shows "Downloading…" while fetching, not "Loading…".
- [x] When audio downloads in the background, the cell icons change to playable
      without refreshing.
- [x] Click "Download" in a cell, go back to the list, and confirm that cell shows
      the downloading spinner, then becomes playable.
- [x] Deleting a recording still clears it and updates the icon.
- [x] Recording, saving, and playing audio still work normally.
- [x] Downloading many files at once doesn't make the screen jitter.


LeviXIII (approved):
Works well between syncs and updates correctly.

PR #1073, opened 2026-06-26
Fix #1038: cancel in-flight chapter-video downloads on delete/replace

### LINKED ISSUE #1038: Bug: Deleted video doesn't stop loading
## Description

The video player begins loading a video, but if the video is deleted while it is still loading, the load operation continues in the background. The loading process should be cancelled/aborted as soon as the video is deleted.

## Steps to Reproduce

1. Open a cell that has a video attached (or start loading/streaming a video).
2. While the video is still loading (before it finishes buffering/loading), delete the video.
3. Observe the network/loading activity.

## Expected Behavior

When the video is deleted, any in-progress loading of that video should immediately stop. No further loading, buffering, or network requests for the deleted video should occur.

## Actual Behavior

The video continues to load even after it has been deleted. The load operation is not cancelled and proceeds to completion in the background.

## Acceptance Criteria

- [ ] Deleting a video while it is loading cancels the in-progress load.
- [ ] No further network/buffering activity occurs for a video once it has been deleted.
- [ ] The behavior works for both locally attached videos and streamed videos (URLs).
- [ ] No console errors or dangling references remain after the load is cancelled.

### Core Scenario (Repro Path)
- [ ] Open a cell with a video attached and begin loading it.
- [ ] While the video is still loading (before it finishes buffering), delete the video.
- [ ] Confirm the in-progress load stops immediately.
- [ ] Confirm no further network/buffering requests occur for the deleted video (verify via network panel/devtools).

### Video Source Variants
- [ ] **Locally attached** video → load cancels on delete.
- [ ] **Streamed video (URL)** → load/stream cancels on delete; no further range/segment requests.
- [ ] Large video that loads slowly (clear loading window) → cancellation observable mid-load.
- [ ] Small/fast video → no errors if delete happens to land right as load finishes.

### Timing & Lifecycle
- [ ] Delete at the very start of loading (before any data buffered) → cancels cleanly.
- [ ] Delete partway through buffering → cancels, partial buffer discarded.
- [ ] Delete just as loading completes → no dangling playback/buffering, no errors.
- [ ] Rapidly start-load-then-delete multiple times → no leaked requests accumulate.

### Post-Delete State
- [ ] Video UI updates to reflect the deletion (player removed / replaced with empty/attach state).
- [ ] No stale/partial video remains playable after deletion.
- [ ] Re-attaching a new video to the same cell loads correctly (the cancelled load doesn't interfere).

### Scope / Targeting
- [ ] Deleting video on one cell does **not** cancel a loading video on a different cell.
- [ ] With multiple cells loading video, only the deleted video's load is aborted.
- [ ] Switching cells while a video loads, then deleting → correct video's load is cancelled.

### Cleanup & Stability
- [ ] No console errors or warnings after cancellation.
- [ ] No dangling references / listeners / abort controllers left behind (no memory leak).
- [ ] No unhandled promise rejections from the aborted load.
- [ ] Closing the cell editor while a video is loading also stops the load (no orphaned background load).

### Regression
- [ ] Normal video load → play → completes works when no deletion occurs.
- [ ] Normal video deletion when **not** loading still works as before.
- [ ] Audio loading/playback in the same cell unaffected by video load cancellation.
- [ ] Sync behavior unaffected by the cancellation logic.

sampjvv:
Closes #1038.

## Problem
While a chapter video is downloading ("Get video → Load/Save to project"), deleting or replacing it doesn't stop the download. The fetch runs to completion in the background and writes the bytes to `files/`, **resurrecting the video the user just deleted**.

## Root cause
`downloadVideoToProject` / `downloadVideoToSessionCache` (`videoDownloadUtils.ts`) fetch the whole LFS object via `authApi.downloadLFSFile` with **no `AbortController`**, then `writeFile` **unconditionally**. The in-flight tracker was a plain `Set` (no way to cancel), the delete/replace handlers never signalled it, and the download progress was `cancellable: false`.

## Fix
- In-flight tracker → `Map<string, AbortController>`. `beginVideoOperation` returns the controller; new `abortVideoOperation()` cancels it.
- Both downloaders accept a `signal`, pass it to `downloadLFSFile` (its type already accepts one), and **guard the write** with `signal.aborted` — so even if the downloader ignores the signal, the deleted file is never rewritten.
- `deleteVideoFile` and the metadata **replace** path call `abortVideoOperation(...)` before removing the file.
- The "Saving/Loading video…" progress is now **cancellable** and wired to abort (a user cancel no longer pops an error).

## Scope decision
The streamed `<video>` path is intentionally unchanged. It **self-limits** — the browser buffers a chunk, goes idle, and releases on GC — so there's no equivalent background-load leak to fix there. (Verified by instrumenting the element's `networkState`/`buffered` live.)

## Verification
- **Unit (6):** the cancellation contract — abort propagates to the signal, re-begin supersedes without leaking, ops scoped per workspace+video.
- **Integration (3):** drives the real `downloadVideoToProject` against a temp workspace + LFS pointer with a mocked, abortable `downloadLFSFile`:
  - control — no abort → bytes are written (harness is genuine);
  - **delete + abort mid-download → file stays deleted, even when the downloader ignores the signal** (the core guarantee);
  - signal-honoring downloader → also stays deleted.
- **Frontier source confirmed** (`genesis-ai-dev/frontier-authentication`): `downloadLFSFile` forwards the `AbortSignal` through `downloadLFSObject` (refcounted) into the underlying `fetch`, so the network request is genuinely cancelled — a shared object download stops once the last waiter aborts.
- **Live GUI:** in a synced, stream-only project, started "Get video → Save to project" and deleted the video mid-download — the file was **not** resurrected.
- `tsc --noEmit` + `eslint` clean. The targeted `videoDownloadUtils` suite is 9/9.

> Note: the full local extension-test run currently hangs in an **unrelated, pre-existing** E2E test (`E2E: Timer Management / Timer During Sync`, project-swap timers) that runs before — and independently of — these video tests; the same changes passed the full suite (1202 passing / 0 failing) before rebasing onto current `main`. CI runs the authoritative suite.


LeviXIII (commented):
There are a couple of places and scenarios where the cancel button is missing. If we try to save or load the video to the project from pressing the Show Video button, the cancel button is missing in that flow. There is also the scenario where the video is missing. If we try to download it, it will try to retrieve it, but we should be able to cancel from there as well.

sampjvv:
@LeviXIII Both missing-cancel scenarios are fixed in f69fe30c. They turned out to share one handler: the Show Video → "Save to project"/"Load for session" buttons and the missing-video "Download & play" button all post `downloadVideoFile`, whose progress toast was the one video download still `cancellable: false`.

- That progress is now **cancellable**, and the fetch is registered in the same in-flight registry as the sidebar flow — so a cancel *or* deleting/replacing the video mid-download aborts it (the delete-abort previously only covered the sidebar downloads). Cancel resets the player quietly ("Download cancelled." + Retry), no error toast.
- Also sealed the write race: a cancel that lands while the verified bytes are being *written* now reverts `files/` to the exact pre-download pointer stub (or removes the file/cache entry if none existed). The abort check brackets the write in all three write sites (handler + both nav utils), and it's the last statement inside the progress callback — so any cancel that can physically be clicked is honored; no window where a cancel is silently ignored and a video stays saved.
- Two new deterministic write-race tests (abort fired during the fs write): stub restored byte-identical / file removed when no stub pre-existed. 11/11 videoDownloadUtils tests green.

Verified in the app: stream-only → Show Video → Save to project, and stream-and-save → Download & play — both show Cancel; cancelling mid-download leaves `files/` as the pointer stub.

LeviXIII (approved):

PR #1074, opened 2026-06-27
Fix #1007: don't export empty audio folders; trim + gate the warning modal

### LINKED ISSUE #1007: We shouldn't export empty folders when exporting
## Current Behaviour
When exporting with the wizard, in step 1 if you select a text only file and then a text + audio option, and then select an audio exporting option in step 2, you will be met with a warning saying that you will have an empty audio folder when exporting the text only file. When exporting after finishing step 3, you will have a folder that has a single NOTICE.txt file that just explains why it exists.

## Acceptance Criteria
- [x] Remove the warning about empty folders in the modal that pops up when selecting an audio exporting option in step 2 of the wizard.
- [x] Change the exporting code to not export an empty audio file and thus, not give you the NOTICE.txt file.

### Warning Removal (Step 2 modal)
- [x] In Step 1, select **a text-only file together with a text+audio file**, advance to Step 2, and choose an **audio export option** → confirm the empty-folder warning modal **no longer appears**.
- [x] Same setup but choose **audio + timestamps** → confirm no warning modal appears.
- [x] Same setup but choose a **non-audio** export format → behavior unchanged (no spurious modals introduced).
- [x] Select **only text-only files** (no audio cells anywhere) and make sure you still cannot pick an audio export option.
- [x] Select **only text+audio files** and pick an audio export option → wizard advances normally, no modal.
- [x] Verify the wizard's **"Next Step"** flow from Step 2 → Step 3 (or → Step 4 for non-Bible groups) works without the modal blocking advancement.

### NOTICE.txt / Empty Audio Folder Removal
- [x] Export a mix of **text-only + text+audio** files using an audio format → confirm the output directory contains **no folder for the text-only file** (or, if a folder exists, it is **not empty** and does **not** contain `NOTICE.txt`).
- [x] Confirm **no `NOTICE.txt` file** is written anywhere in the export output.
- [x] Export only **text+audio** files → audio folders for those files are populated normally (regression check).
- [x] Export **text+audio** files where **some cells have no recording** → the file's audio folder still exists with recorded cells; per-cell "no audio recorded" reporting still works as before.
- [x] Export a file that has audio attachments but where every cell is filtered out (e.g. by milestone selection excluding all milestones with audio) → confirm no empty folder + NOTICE.txt is left behind.
- [x] Combined audio + non-audio bundle (when export format produces both) → text-only files still appear correctly in the non-audio output; only the audio side omits them.

### Regression Testing
- [x] Export progress reporter still completes cleanly (no stalled "preparing" / missing-file count mismatch caused by skipping the empty folder).
- [x] Wizard step indicator and Back/Next buttons behave identically to before for all non-affected flows.
- [x] Re-running an export to the same destination still overwrites/refreshes correctly and does not leave orphaned `NOTICE.txt` from a previous run (worth checking if you're testing on top of a folder that previously had one).

Image

Image

Image

sampjvv:
Closes #1007.

## Problem
On an **audio** export that includes a file/chapter with no audio: the wizard wrote an empty per-book folder containing only a `NOTICE.txt`, and the step-2 "Files Without Audio" warning could be dismissed (X / click-outside) **without** acknowledging it.

Testing also surfaced an inconsistency: a *book* is one `.codex` file and *chapters* are **milestones** within it. The warning/NOTICE logic was per-file, but the exporter silently drops per-chapter — so an empty chapter inside an audio-bearing book vanished with **no feedback at all**.

## Changes
**1. Trim the warning modal** (`projectExportView.ts`) — drop the now-inaccurate "…their exported audio folders will be empty" / "…will produce empty output…" wording from the shared warning popup; keep the header + "The following files have no audio/text translations." + file list.

**2. Stop writing the empty folder + `NOTICE.txt`** (`exportHandler.ts`) — delete `generateMissingContentNotices()` (the only code creating the per-book folder + `NOTICE.txt`), its call site, and the now-unused `analyzeNotebookContent` import. `reporter.complete()` is unchanged.

**3. Report skips in the completion summary instead — file *and* chapter level** (`audioExporter.ts`) — per book, track which selected milestones produced audio. A whole no-audio book → `"Skipped (no audio): N book(s) — …"`; selected-but-empty chapters → `"Chapters skipped (no audio): GEN 1; …"` (book-prefixed, capped at 6). These flow through the existing `extraMessages` → completion summary, so there's **no new plumbing**. Only chapters that were *selected but produced no audio task* count (download failures stay in their existing category; the whole-book opt-out is excluded).

**4. Require explicit OK to dismiss the warning** (`projectExportView.ts`) — the warning is a full-screen blocking modal; removing the **X** button and the click-outside dismissal leaves **OK** as the only way out, so the export can't proceed until it's acknowledged. Applies to both the audio and text warnings (shared popup). No keyboard/Escape bypass exists for this popup.

## Verification
- `tsc --noEmit` + `eslint` + `npm run compile` clean.
- New tests (`src/test/suite/export/audioSkipSummary.test.ts`): **book-level** (no `NOTICE.txt` written + skip line in the summary) and **chapter-level** (an empty selected chapter is reported as skipped, not silently dropped). Existing `exportCancellation` test still passes — 3 export tests green.
- Manual: (a) mixed text-only + text+audio export → trimmed modal, no `NOTICE.txt`, skipped books/chapters listed in the summary; (b) the warning modal has no X and ignores click-outside — only OK dismisses it.

> Note: the warning-modal changes are in the extension's wizard HTML string (built by `npm run compile`), which has no DOM test harness in the suite — verified manually.


sampjvv:
### Follow-up edge cases — "Files Without Audio" warning (out of scope for this PR)

Flagging for @LeviXIII to look into. The no-audio warning only fires the moment you click the *Include Audio* option, so it can be bypassed when the selection changes afterward. Two cases it misses:

1. **Deselecting the audio chapters → no warning.** Export a book that *has* audio (no warning, correct). Then on the chapter-selection step, deselect the only chapter(s) that actually contain audio. The remaining selection has no audio, but you get no warning — you just export "0 audio files." → Should warn, on the chapter step, when the current selection contains no audio at all.

2. **Re-selecting files after the format is set → no warning.** Pick a book with audio, choose Include Audio (warning logic runs, nothing to flag), advance. Go back and add a *no-audio* book to the selection. Because the audio format is already remembered, you can advance straight through with no warning about the no-audio book. → Should re-check (and warn) whenever the *file* selection changes, not just on the audio-option click.

**Common thread:** the warning is a one-shot trigger; it should re-evaluate whenever the file **or** chapter selection changes (i.e. gate on "current selection has no audio," re-checked at each step).

sampjvv:
Update: scenario 2 (re-selecting files after the audio format is already set) is now fixed in this PR (53c7aba4) — arriving at step 2 from the file step re-runs the no-audio check when the audio option is still selected and the file selection changed. An unchanged, already-acknowledged selection is not re-warned. Scenario 1 (chapter-level deselection) remains open.

LeviXIII (approved):

PR #1088, opened 2026-07-07
Add Resolve function to Sparkle process

### LINKED ISSUE #1071: Build HTML enforcement into Sparkle button
## Summary
When a user clicks the **Sparkle** (AI translation) button on a cell, the editor should automatically insert any missing structural HTML tags as part of the same action — so the translated cell comes back with a correct, source-matching HTML structure without any extra clicks.

Today this auto-resolution is only available for IDML-imported content, where the round-trip back to InDesign requires the cell's HTML structure to exactly match the source. The Sparkle flow needs to fold the existing **Resolve** step into itself for these cells.

## Problem

The IDML importer now relies on the resolve step to insert the correct HTML tags into a cell so the structure matches the source. This is required for a clean IDML round-trip export back to InDesign, where non-translatable structural elements (formatting tags, spans, paragraph/heading tags, line breaks, USFM markers) must be preserved exactly.

**Current workflow:** When a user clicks Sparkle, the LLM translates the text but frequently drops or alters the structural HTML. The editor detects the mismatch (`compareHtmlStructure`), glows the cell yellow, shows a "Structure mismatch" banner, and surfaces a **Resolve** button. The user then has to manually click **Resolve** to trigger a *second* LLM pass (`resolveHtmlStructure`) that re-inserts the missing tags.

**Why this is a problem:** It is a manual two-step process. A user can finish translating with Sparkle, never notice the mismatch warning, and end up with cells that break the IDML round-trip export. The structure correction should happen automatically, not depend on the user spotting a warning and clicking a second button.

## Acceptance Criteria
- [ ] Clicking the Sparkle button automatically runs the HTML-structure resolution step after translation completes — no manual "Resolve" click required.
- [ ] After the automatic resolve, the cell's HTML structure matches the source skeleton (`compareHtmlStructure` returns `isMatch: true`), with all translated text preserved unchanged.
- [ ] The structure mismatch warning (yellow glow / banner) does **not** appear for cells successfully auto-resolved via Sparkle.
- [ ] The loading/spinner state remains active through both the translation and the auto-resolve passes; it only clears when the cell is fully resolved.
- [ ] If the automatic resolve fails (LLM error, offline, still mismatched), the cell falls back to the current behavior: the mismatch warning and manual **Resolve** button are shown so the user can retry.
- [ ] No translated wording is reverted to the source language during the auto-resolve step (only structural tags are added).
- [ ] Auto-resolution is scoped correctly — it only runs where HTML-structure enforcement applies (e.g. IDML / `enforceHtmlStructure` cells) and does not run on source cells.
- [ ] The manual **Resolve** button remains available as a fallback for cells that still have a mismatch after the automatic attempt.
- [ ] Resolved content is persisted to the document via the existing `updateCellContent` path (survives reload).

## Test Checklist
- [ ] Import an IDML source file and confirm cells carry the structure metadata used for resolution.
- [ ] Click Sparkle on a target cell whose translation drops structural tags → verify tags are auto-inserted and the structure matches the source with no manual click.
- [ ] Confirm the mismatch warning/glow and the "Structure mismatch" banner do not appear after a successful auto-resolve.
- [ ] Verify the spinner stays visible for the full translate → auto-resolve sequence and clears only at the end.
- [ ] Verify translated text content is unchanged after auto-resolve (only tags added; no words reverted to source language).
- [ ] Simulate a resolve failure (offline / forced LLM error) → confirm fallback to the manual Resolve button and warning state.
- [ ] Click Sparkle on a cell whose translation already matches the source structure → confirm no unnecessary second pass / no errors.
- [ ] Confirm auto-resolve does not trigger on source-text cells or in projects without `enforceHtmlStructure`.
- [ ] Reload the editor after auto-resolve and confirm the resolved HTML persists.
- [ ] Run an IDML round-trip export with auto-resolved cells and confirm the exported `.idml` opens cleanly in InDesign with correct structure.
- [ ] Click the manual **Resolve** button on a cell still mismatched after auto-resolve → confirm it still works as before (regression check).

Fikitti:
When sparkling and the HTML enforcement is on, the AI translation will finish and before writing into cell, it will also trigger the "Resolve" function to enforce all the missing html tags to preserve the styles. The original "Resolve" button is still there as a fallback function.

## PR Title
Format:
Add-Resolve-function-to-Sparkle-process-1071

## Summary
Closes #1071

The sparkle button has 2 steps now, also shown in the loading bar on the bottom of the screen. First is the AI translating the content and second is the HTML enforcement that will resolve the cell tag missmatch before the final translation is written in the cell.

## Changes

## Test Checklist
- [x] Import style heavy file format, like biblica study bible, with HTML enforcement enabled.
- [x] Try sprakle button on cells, they should come out always green without the tag missmatch warning.
- [x] Some might have the warning still if they contain complicated tags or lot of them, and the "Resolve" button is still there as a fallback feature.
- [x] Later try exporting it with round-trip export and see if the styles were preserved.

### DIFF
--- a/src/projectManager/utils/migrationUtils.ts
+++ b/src/projectManager/utils/migrationUtils.ts
@@ -3594,7 +3594,10 @@ export const migration_repairVerseRangeDuplication = async (): Promise<void> => 
                         const report = await repairVerseRangeDuplicationForFile(file, {
                             dryRun: true,
                         });
-                        if (report.changed && report.tombstoned > 0) {
+                        // Include reposition-only files (tombstoned === 0): duplicates may already
+                        // be soft-deleted while their live counterparts are still stranded out of
+                        // their chapter (e.g. after a sync merge reverted an earlier repair).
+                        if (report.changed) {
                             affected.push({ uri: file, tombstoned: report.tombstoned });
                             totalTombstones += report.tombstoned;
                         }
@@ -3648,7 +3651,10 @@ export const migration_repairVerseRangeDuplication = async (): Promise<void> => 
         const previewLimit = 10;
         const previewLines = affected
             .slice(0, previewLimit)
-            .map((r) => `  • ${path.basename(r.uri.fsPath)} (${r.tombstoned})`)
+            .map(
+                (r) =>
+                    `  • ${path.basename(r.uri.fsPath)} (${r.tombstoned > 0 ? r.tombstoned : "reposition only"})`
+            )
             .join("\n");
         const moreLine =
             affected.length > previewLimit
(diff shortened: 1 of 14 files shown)


LeviXIII (commented):
1. Works well for the most-part. There is a slight misalignment when I tested with a .docx file that had stylings on a round-trip. I'll post the screenshots below.

Screenshot 2026-07-08 at 12 40 55 PM

2. I ran the change in a project and noticed that it did resolve the cells correctly. However, the users would have to click "resolve" one by one in order to fix what's already been translated and then have to re-validate. Is there a way that we can just create a script to do that so they don't have to?

LeviXIII (approved):
The .docx issue is known and will be addressed at a later date. This is working well for resolving cells and even allows for resolving a group of cells at once.

PR #1089, opened 2026-07-08
1075-patani-malay-project-formatting-issues-2

### LINKED ISSUE #1075: Patani malay project: Formatting Issues
## Description
Books in the Pattani Malay project show verses that look merged together, recent translations that appear to be missing (with no normal edit history), and previously removed headings/paratext that have come back.

**Root cause (found during investigation):** the affected books ended up stored in *two versification layouts at the same time* — combined verse-range cells (e.g. one cell for `MAT 8:14-15`) and separate single-verse cells (`MAT 8:14`, `MAT 8:15`) — living side by side. During git sync, the merge matches cells by their internal ID; because the two layouts have different IDs, it keeps **both** instead of reconciling them, and stray empty copies pile up at the end of the file. The translated text ends up split across the two copies (sometimes in the range cell, sometimes in the singles), so a verse can look "merged" or look empty/"missing" depending on which copy the editor shows. The sync-time changes were written as system "migration" edits rather than normal user edits, which is why recent work looks absent from history. Removed headings reappear because sync re-adds cells that still exist on another translator's device.

This began when the verse-range migration/reordering started running during sync (shipped in **v0.28.0**); a later fix (**v0.29.0**) only partially addressed it. The full fix is to reconcile the duplicate layouts during merge (so it can't recur) plus a one-time repair for already-affected projects.

**Note:** the separate problem where user "pmsue" cannot update the app and sees an error is a client/version (binary) issue, **not** part of this data bug. It is listed here for completeness but needs its own resolution.

## Steps to Reproduce
1. Open project and navigate to Mt 8:14-15; 9:1-2; 13:1-2, 24-25, 34-35, 45-46; 14:6-7, 34-35; 15:5-6.
2. Observe verses merged into single cells and missing recent translations with no corresponding edit history; check Mark 1 before v21 for reappeared test headings/paratext.

## Expected Behavior
- Versification matches the source (each verse remains in its own cell).
- Removed headings/paratext do not reappear.
- No silent deletions; all changes appear in edit history.

## Actual Behavior
- Separate verses are merged into combined cells (two layouts of the same verses coexist).
- Some recent translations appear missing and are not shown in normal edit history.
- Previously removed headings/paratext reappear after sync.

## Acceptance Criteria
- [ ] Verses display in separate cells matching source versification.
- [ ] Removed headings/paratext do not reappear.
- [ ] No silent data loss; deletions/changes are tracked in edit history.
- [ ] Identify and fix root cause; prevent recurrence.
- [ ] Restore affected passages where feasible.
- [ ] Provide steps for users to check the app's binary version.

## Test Checklist
- [ ] Repair command collapses each duplicated verse to a single cell, translated text preserved.
- [ ] The repair never deletes a cell that contains text (only empty duplicates are removed).
- [ ] Re-running the repair makes no further changes (idempotent).
- [ ] After a sync, the duplicates do not come back (merge reconciles the two layouts).
- [ ] Cases where both layouts contain text (e.g. Mt 12:46-47) are flagged for manual review, not
      auto-merged.
- [ ] Previously removed headings/paratext stay removed after a sync.
- [ ] Document/subtitle (non-scripture) projects are unaffected by the repair and merge change.
- [ ] Affected Pattani Malay passages (Mt 8, 9, 13, 14, 15; Mark 1) verified in the editor.

Image

Image

Image

Interface URL: https://airtable.com/appfEUXIPganSIP89/pagyewiWtbGAM67M4/recTbYrjKO9GEXT1L?home=pagy8X7KblenRWOQD

LeviXIII:
## Summary
Closes #1075

The `Codex: Repair Duplicated Verse Cells` command silently skipped files that only needed repositioning. Its scan gated on `report.changed && report.tombstoned > 0`, so a file whose empty duplicate cells were *already* soft-deleted — but whose live, translated verse cells were still stranded at the end of the file (e.g. after a sync merge reverted an earlier repair's ordering, as happened to the Patani Malay project's MAT.codex) — was dropped from the apply list and the command reported "No duplicated verse cells detected."

This PR makes the scan include reposition-only files and adjusts the command's messaging so that case is described accurately.

## Changes
- `migration_repairVerseRangeDuplication` scan now gates on `report.changed` alone, so files needing only the reposition pass (zero new tombstones) are offered for repair.
- Confirm dialog: reposition-only files are listed as "(reposition only)" instead of "(0)", and when no duplicates exist the headline reads "Found N file(s) with verse cells stranded out of their chapter" instead of "Found 0 empty duplicate verse cell(s)".
- Confirm dialog detail now mentions that stranded verse cells are moved back to their chapter.
- Completion toast says "stranded cells repositioned" instead of the misleading "0 duplicate(s) removed" when no tombstones were created.
- No changes to the repair logic itself (`repairVerseRangeDuplicationForFile`, `planVerseDuplicationRepair`, `reorderVerseRangeCells` are untouched).

### Scan gate (verified against the real affected file)
- [x] Dry-run on the Pattani Malay `MAT.codex` returns `tombstoneIds: 0, orderChanged: true` — the old gate excluded it, the new gate includes it
- [x] Files with neither duplicates nor stranded cells are still excluded (`changed: false`), so the command remains a no-op on healthy projects

### Apply path (verified against the real affected file)
- [x] All 1179 cells and cell IDs preserved after apply
- [x] All 22 stranded verses (MAT 12:46–15:6 translations, 17:21, 22:34–35, 22:41–42, 24:35) end up under their correct chapter milestone
- [x] Stranded translations survive intact (e.g. 13:1 and 13:2 keep pmsue's text)
- [x] Command re-run end-to-end from the extension host on the affected project: MAT.codex is offered and repaired correctly

### Messaging
- [x] Reposition-only path shows the stranded-cells headline and "(reposition only)" file labels, and completes with "stranded cells repositioned"

### DIFF
--- a/src/projectManager/utils/migrationUtils.ts
+++ b/src/projectManager/utils/migrationUtils.ts
@@ -3594,7 +3594,10 @@ export const migration_repairVerseRangeDuplication = async (): Promise<void> => 
                         const report = await repairVerseRangeDuplicationForFile(file, {
                             dryRun: true,
                         });
-                        if (report.changed && report.tombstoned > 0) {
+                        // Include reposition-only files (tombstoned === 0): duplicates may already
+                        // be soft-deleted while their live counterparts are still stranded out of
+                        // their chapter (e.g. after a sync merge reverted an earlier repair).
+                        if (report.changed) {
                             affected.push({ uri: file, tombstoned: report.tombstoned });
                             totalTombstones += report.tombstoned;
                         }
@@ -3648,7 +3651,10 @@ export const migration_repairVerseRangeDuplication = async (): Promise<void> => 
         const previewLimit = 10;
         const previewLines = affected
             .slice(0, previewLimit)
-            .map((r) => `  • ${path.basename(r.uri.fsPath)} (${r.tombstoned})`)
+            .map(
+                (r) =>
+                    `  • ${path.basename(r.uri.fsPath)} (${r.tombstoned > 0 ? r.tombstoned : "reposition only"})`
+            )
             .join("\n");
         const moreLine =
             affected.length > previewLimit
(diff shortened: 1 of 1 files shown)


TimRl (approved):
Looks good!

PR #1093, opened 2026-07-10
1064-algerian_arq-target-text-and-source-have-varying-ranges

### LINKED ISSUE #1064: Algerian_arq: Target text and source have varying ranges
- Problem: A mismatch between source and target line ranges in episode 208 (source ends at 714, target continues to 835 with non-clickable pages and child cells) is blocking alignment and navigation for the translator.
- Context: In the Algerian_arq_The_Chosen project, episode 208’s source correctly ends at line 714, but the target extends to 835, with the last two pages not clickable and child cells appearing after 714. This discrepancy is slowing translation and disrupting normal workflow.

## Description
The "add child cell" (split) button in the Codex Cell Editor creates **duplicate child cells with identical timestamps** when the same cell is split more than once in an editing session. Instead of each split advancing through the parent cell's time range, repeated splits all carve off the *same* `[midpoint, endTime]` slice, producing overlapping duplicate children (same time range, same/typed content, different IDs).

This surfaced as "random child cells" appearing in subtitle/dubbing projects after the late‑May deploy. 

**Root cause:** `makeChild` in `webviews/codex-webviews/src/CodexCellEditor/TextCellEditor.tsx` (~line 2246) computes the child range from the **prop** `cellTimestamps` (line ~2250) rather than the freshest/effective timestamps. That prop does not refresh after a split within the same open editor, so:
- `childStartTime = startTime + (endTime - startTime) / 2` uses the **original** range on every click.
- The parent's `endTime` update becomes a no‑op the 2nd/3rd time (same value), confirming staleness.

## Steps to Reproduce
1. Open a cell that has start/end timestamps (e.g. a subtitle/dubbing cell) in the Codex Cell Editor.
2. Click the **add child cell** (split) button; type text into the new child.
3. Without closing/reopening the parent, click the **add child cell** button again on the same parent cell.
4. (Optional) Repeat once more.

## Expected Behavior
Each split should operate on the parent's **current** time range, carving off the next sub‑segment so the resulting children are **contiguous and non‑overlapping** (e.g. `[a, m1]`, `[m1+, m2]`, `[m2+, end]`). Repeated splits advance through the cell.

## Actual Behavior
Every repeated split produces a child with the **identical** `[midpoint, endTime]` range as the previous one, because `makeChild` reads a stale `cellTimestamps`. Result: 2–3 overlapping duplicate child cells with the same timestamps (and, after typing, often identical content), e.g. `TheChosen_106` parent `75f3c8fe` has three children all at `[1559.6845, 1561.228]` while the parent has only one `endTime` edit.

## Acceptance Criteria
- [ ] `makeChild` derives the child range from the freshest/effective parent timestamps (reflecting prior splits), e.g. `effectiveTimestamps` plus an optimistic local update of the parent's new `endTime`, or by re‑reading the live cell data after the update round‑trips.
- [ ] Splitting the same cell N times yields N contiguous, non‑overlapping children — no two siblings share a time range.
- [ ] A guard prevents creating a child whose range duplicates an existing sibling or whose duration is ≤ a minimum threshold (no zero/near‑zero or overlapping segments).
- [ ] No regression to the single‑split path or to paratext creation (`addParatextCell`), which has the same `cellTimestamps`‑prop pattern.
- [ ] Fix is in the webview split logic only; the merge resolver is unchanged.

## Test Checklist
- [ ] Unit/component test: two consecutive `makeChild` calls on one cell produce **distinct, contiguous, non‑overlapping** time ranges.
- [ ] Unit/component test: three consecutive splits produce three advancing segments covering the original range with no overlap.
- [ ] Test: duplicate‑range guard rejects/no‑ops a split that would reproduce an existing sibling's range.
- [ ] Test: minimum‑duration guard prevents creating zero/near‑zero‑length children.
- [ ] Test: the parent's `endTime` is correctly updated and reflected before the next split (no stale‑prop reuse).
- [ ] Regression: single split still yields parent `[start, mid-0.001]` + child `[mid, end]`.
- [ ] Regression: `addParatextCell` paratext creation still behaves correctly.
- [ ] Manual: split a subtitle cell repeatedly in a dubbing project; confirm no identical‑timestamp duplicate children appear.
- [ ] Manual: split, sync between two peers, and confirm no duplicate children are introduced or resurrected by merge.

Image

Image

- Interface URL: https://airtable.com/appfEUXIPganSIP89/pagyewiWtbGAM67M4/rec4hcFdgi5P6wK1o?home=pagy8X7KblenRWOQD

LeviXIII:
## Summary

Closes #1064

In files where cells have been split, the milestone dropdown showed extra pages past the end of the content. Episode 208's lines end at 714, but the dropdown advertised 17 pages instead of 15; clicking the phantom pages just showed the last real page.

Root cause: `buildMilestoneIndex` counted child cells (splits) in `milestone.cellCount`, while `getCellsForMilestone` paginates by root cells only — children display on their parent's page. The webview derives the dropdown's page list from `cellCount`, so the inflated count produced pages that don't exist. The fix counts root cells only, matching the pagination, subsection progress, and line-numbering logic (which all already excluded children).

## Changes
- `buildMilestoneIndex` excludes cells with a `parentId` (checked in both `metadata.parentId` and `metadata.data.parentId`) from `milestone.cellCount` and `totalCells`; child cells still get `milestoneIndex` assigned
- Milestone dropdown page count and page labels (e.g. "701-714") now match the visible line numbers, which skip child cells
- Cross-milestone line-number offsets (`calculateLineNumberOffset`) are correct in split files, since they're derived from the same count
- Added 5 regression tests to `milestonePagination.test.ts` covering root-only counting, dropdown/pagination agreement, deleted children, navigation to child/paratext cells on later pages, and subsection progress

### Dropdown page count
- [x] Dropdown page count derived from `cellCount` matches `getSubsectionCountForMilestone` in a split document (new test)
- [x] Deleted cells and deleted children are excluded from the count (new test)
- [x] Files without splits are unaffected

### Child and paratext cells
- [x] Creating child cells still works 
- [x] Child and paratext cells still display on their parent's page (existing + new tests)
- [x] Navigating to a child or paratext cell on a later page resolves to the correct, dropdown-reachable subsection (new `findMilestoneAndSubsectionForCell` test)

### Progress
- [x] Subsection progress produces exactly one entry per dropdown page, and children with content don't skew percentages (new `calculateSubsectionProgress` test)
- [x] Milestone-level progress unchanged (existing `calculateMilestoneProgress` tests pass; progress always excluded children via `shouldExcludeCellFromProgress`)

### DIFF
--- a/src/providers/codexCellEditorProvider/codexDocument.ts
+++ b/src/providers/codexCellEditorProvider/codexDocument.ts
@@ -1622,7 +1622,12 @@ export class CodexCellDocument implements vscode.CustomDocument {
             // Process content cells (excluding milestones, paratext, and deleted)
             if (cellType !== CodexCellTypes.MILESTONE && cellType !== "paratext") {
                 const isDeleted = cell.metadata?.data?.deleted === true;
-                if (!isDeleted) {
+                // Child cells (splits) are excluded from counts: pagination, subsection labels,
+                // and line numbering are all root-based (children display on their parent's page)
+                const isChildCell =
+                    cell.metadata?.parentId !== undefined ||
+                    (cell.metadata?.data as { parentId?: string; } | undefined)?.parentId !== undefined;
+                if (!isDeleted && !isChildCell) {
                     totalContentCells++;
                 }
 
@@ -1636,8 +1641,8 @@ export class CodexCellDocument implements vscode.CustomDocument {
                         cell.metadata.data = {} as any;
                     }
                     (cell.metadata.data as any).milestoneIndex = currentMilestoneIndex;
-                    // Only count non-deleted cells for milestone cell count
-                    if (!isDeleted) {
+                    // Only count non-deleted root cells for milestone cell count
+                    if (!isDeleted && !isChildCell) {
                         currentMilestoneCellCount++;
                     }
                 }
--- a/src/test/suite/milestonePagination.test.ts
+++ b/src/test/suite/milestonePagination.test.ts
@@ -182,6 +182,175 @@ suite("Milestone-Based Pagination Test Suite", () => {
         assert.strictEqual(milestoneIndex.totalCells, 1, "Total cells should exclude paratext");
     });
 
+    test("buildMilestoneIndex excludes child cells from count (root-based pagination)", async () => {
+        const cells = [
+            {
+                kind: 2,
+                languageId: "scripture",
+                value: "1",
+                metadata: {
+                    type: CodexCellTypes.MILESTONE,
+                    id: "milestone-1",
+                },
+            },
+            {
+                kind: 2,
+                languageId: "scripture",
+                value: "Root 1",
+                metadata: {
+                    type: CodexCellTypes.TEXT,
+                    id: "GEN 1:1",
+                },
+            },
+            {
+                kind: 2,
+                languageId: "scripture",
+                value: "Child via metadata.parentId",
+                metadata: {
+                    type: CodexCellTypes.TEXT,
+                    id: "GEN 1:1:child-1",
+                    parentId: "GEN 1:1",
+                },
+            },
+            {
+                kind: 2,
+                languageId: "scripture",
+                value: "Child via data.parentId",
+                metadata: {
+                    type: CodexCellTypes.TEXT,
+                    id: "GEN 1:1:child-2",
+                    data: {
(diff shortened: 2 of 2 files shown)


TimRl (approved):

TimRl:
Looks good and it works!

Issue #1107, opened 2026-07-22
Bug: Structure mismatch errors on docx-imported projects: spurious warnings, broken "Resolve", and lost formatting


TimRl:
# Structure mismatch errors on docx-imported projects: spurious warnings, broken "Resolve", and lost formatting

## Summary

Projects imported from `.docx` (with `enforceHtmlStructure` enabled) showed persistent
**"Structure mismatch"** warnings on translated cells. Clicking **Resolve** often failed to fix
them — and in the worst case **replaced the translation with the English source text**. Even when
a translation was generated with the correct structure, saving the cell from the editor stripped
the source's formatting wrappers and re-introduced the mismatch.

## Symptoms

- Cells show `Structure mismatch: Extra tags: <span>, </span>` immediately after AI translation.
- Cells show `Structure mismatch: Missing tags: <p>, </p>` or `Missing tags: <span>, </span>`
  after saving a cell that was open in the editor.
- Clicking **Resolve** sometimes returns the same mismatch, or re-surfaces source-language
  (English) text inside the translation.
- Blank lines added by the user (Enter presses) triggered mismatch warnings.

## Root causes

There were several independent bugs compounding each other:

1. **LLM completion pipeline wrapped predictions in a bare `<span>`**
   (`llmCompletion.ts`). Every AI translation gained a `<span>` wrapper the source didn't have →
   instant "Extra tags" mismatch.

2. **The docx importer split single text blocks into multiple `<span>` runs.** Word stores
   formatting as runs (`<w:r>`); adjacent runs with identical formatting were each converted to
   their own `<span>`, inflating the source skeleton.

3. **The old "Resolve" prompt only handled *missing* elements** and had no output verification.
   For "extra tags" cases the LLM guessed — sometimes echoing the source text back, overwriting
   the translation with English.

4. **The Quill editor destroys unregistered markup.** Quill silently drops any format it doesn't
   know: `<p style="line-height: 2">` loses its style, and `<span style="font-family: Arial">`
   is removed entirely. Saving a cell that was open in the editor therefore stripped the docx
   wrappers — even overwriting a correctly-structured AI translation one second after it was
   generated.

5. **Several write paths bypassed structure resolution entirely** (A/B test variant selection,
   apply-translation from other views, editor saves).

## Fix

- **Prevention** — content is repaired *before* it lands in the cell, so warnings rarely appear:
  - Removed the spurious `<span>` wrapper from the LLM completion pipeline.
[...]

PR #1108, opened 2026-07-22
Fix html tags issues that aren't resolving as well as translation

TimRl:
## Fix html tags issues that aren't resolving as well as translation
Format:
#1107 

## Summary

Fixes the persistent "Structure mismatch" errors on docx-imported projects (`enforceHtmlStructure`), where AI translations, editor saves, and the Resolve button could all produce or fail to fix mismatches — and Resolve could even overwrite a translation with the source-language text.

The fix works at three levels:

1. **Prevention** — stop mismatches from being created: the LLM pipeline no longer wraps predictions in a spurious `<span>`, the docx importer merges adjacent same-format runs, all LLM write paths route through auto-resolve, and editor saves deterministically restore the styled wrappers that Quill strips (Quill drops any markup it has no registered format for, e.g. `<span style="font-family: Arial">`).
2. **Resolution** — the resolver now tries safe, deterministic fixes first (remove/convert bare `<span>`/`<p>` artifacts, re-dress the target with the source's verbatim wrapper chain). The LLM is only a fallback, with a rewritten prompt that handles both missing *and* extra tags, and its output is verified (structure must match, text must not revert to source language) before saving.
3. **Repair** — a one-shot migration fixes existing projects on activation: applies the deterministic fixes and restores translations the old resolver overwrote with source text (recovered from the cell's edit history, recorded as `MIGRATION` edits).

Also includes small quality-of-life fixes: the structure comparison ignores empty paragraphs (`<p><br></p>` from Enter presses), the Resolve All button only shows when the loaded section actually has mismatches, and the Resolve/Autocomplete modals support Escape-to-close and Enter-to-confirm.

## Changes

- **`sharedUtils/htmlStructureUtils.ts`**
  - `tryDeterministicStructureFix`: verified, text-preserving fixes — re-wrap with the source's verbatim wrapper chain (preserves styles/attributes for round-trip export), remove/convert bare `<span>`/`<p>` pairs, wrap plain text in the source's block wrapper.
  - `compareHtmlStructure` ignores empty paragraphs on both sides.
  - `extractPlainTextFromHtml` for revert-to-source detection.
- **`src/providers/translationSuggestions/llmCompletion.ts`**: removed the automatic `<span>` wrapper around predictions (the root cause of "Extra tags: `<span>`" after every AI translation).
- **`src/providers/codexCellEditorProvider/utils/htmlStructureResolver.ts`**
  - Tiered resolve: deterministic first, LLM fallback, output verified before saving; explicit `StructureResolveOutcome` statuses.
  - `maybeRepairStructureDeterministically`: deterministic-only repair for user saves (never calls the LLM).
- **`codexCellEditorMessagehandling.ts` / `codexCellEditorProvider.ts`**: `saveHtml` repairs Quill-stripped wrappers before writing; A/B variant selection and `applyTranslation` now route through auto-resolve; single and batch resolve use the new outcome statuses with accurate user feedback.
- **`webviews/.../NewSourceUploader/importers/docx/index.ts`**: merges adjacent runs with identical formatting on import (fewer spurious `<span>`s in source skeletons).
- **`src/projectManager/utils/htmlStructureRepairMigration.ts`** (new) + `extension.ts` + `package.json`: one-shot per-workspace migration (`htmlStructureArtifactsRepairCompleted` flag) that repairs artifacts and restores overwritten translations in existing `enforceHtmlStructure` projects.
- **`webviews/.../CodexCellEditor/Editor.tsx` / `TextCellEditor.tsx`**: preserve paragraph structure on save when `enforceHtmlStructure` is on (no more first-`<p>`→`<span>` conversion).
- **`CodexCellEditor.tsx` / `ResolveAllModal.tsx` / `AutocompleteModal.tsx`**: Resolve All button only shown when the loaded section has mismatches; Escape closes both modals; Enter in the Autocomplete number field starts the run.

All behavioral changes are gated on `enforceHtmlStructure` (opt-in, set by the docx importer); legacy USFM/plaintext projects are unaffected. The only ungated change is removing the cosmetic `<span>` wrapper from LLM output.

### Automated (verified)

- [x] Webview unit tests pass (`pnpm test` in `webviews/codex-webviews`), including new suites for the deterministic fixes, re-wrap logic, and empty-paragraph tolerance.
- [x] Extension unit tests added for resolver verification (`verifyResolvedContent`, `maybeAutoResolveHtmlStructure`, `maybeRepairStructureDeterministically`) and migration recovery (`findRevertedTranslation`).
- [x] Extension compiles cleanly (`npm run compile`); all webviews build (`pnpm run build:all`).
- [x] Reproduced the three failure modes against a real docx-imported project (spurious spans after AI translation, resolver reverting text to English, editor save stripping styled wrappers) and confirmed the deterministic fixes handle each cell's actual HTML.

### Manual QA (docx project with `enforceHtmlStructure`)

- [x] AI-translate a cell (sparkle) with the cell open — no mismatch warning after save; styled wrappers intact in the `.codex` file.
- [x] Edit and save a styled cell manually — wrappers restored, translated text unchanged.
- [x] Click Resolve on a mismatched cell — fixes without re-surfacing source-language text.
- [x] Open an existing project with damaged cells — migration notification appears; translations restored, spans unwrapped.
- [ ] Add blank lines with Enter — kept on save, no mismatch warning.
- [x] Escape closes the Resolve All and Autocomplete modals; Enter in the Autocomplete number field starts the run.

### Regression

- [x] USFM/plaintext (non-enforce) project: translate, edit, save — behavior unchanged, no new warnings.
- [x] Docx round-trip export still produces the original formatting.

Screenshot 2026-07-20 at 14 16 40
Screenshot 2026-07-21 at 01 26 46
image


LeviXIII (commented):
Things seem to be working ok except for the line breaks causing the mismatch label to come up. In this case, I think it should warn of that as it should try to remove the extra space to keep the original in line with the new.

I also don't really like the idea of the migration running upon opening a project. I think it should be a manually triggered migration as not all projects were broken. It may have unintended side-effects.

dadukhankevin (approved):
impeccable

Issue #1109, opened 2026-07-22
Bug: Double entries from some word files


TimRl:
# DOCX import duplicates text-box content (`mc:AlternateContent`), corrupting source and target files

## Description

Importing a `.docx` file that contains **floating text boxes** produces duplicated and merged cells in the generated `.source` file (and consequently in the `.codex` target file). The same passage can appear in 3+ consecutive cells, and in some cases two unrelated passages are glued together into a single "Frankenstein" cell.

### Example

Importing `1 Cor L#1 Intro W.docx` produced a source file where these passages each appear **three times** (cells 6–11):

> "Using only the text of 1 Corinthians, try to answer the following questions. …"
> "At this point in the lesson you may not be able to answer every question using only your Bible. …"
> "You will have an opportunity at the end of this lesson to come back and complete your answers…"

and this passage twice more in cells 24–26:

> "Now you can go back and add to your answers above using a Bible dictionary or Bible commentary. …"

## Root cause

Word serializes floating text boxes **twice** inside `w:document.xml` using markup compatibility:

```xml
<w:p>                                 <!-- anchor paragraph -->
  <w:r>
    <mc:AlternateContent>
      <mc:Choice Requires="wps">      <!-- modern Word: drawing + w:txbxContent -->
        ... <w:txbxContent> <w:p>…text…</w:p> </w:txbxContent> ...
      </mc:Choice>
      <mc:Fallback>                   <!-- legacy Word: VML copy of the SAME text -->
        ... <w:txbxContent> <w:p>…text…</w:p> </w:txbxContent> ...
      </mc:Fallback>
    </mc:AlternateContent>
  </w:r>
</w:p>
```

The importer's paragraph enumeration (`extractParagraphXmlList`) used a non-greedy regex over the body XML, which:

1. **Matched nested paragraphs** inside `w:txbxContent` as if they were top-level paragraphs, and
2. **Matched them in both the `mc:Choice` and `mc:Fallback` branches**, importing the same text twice more,
3. Mis-paired `<w:p>` open/close tags across nesting levels, occasionally producing merged cells containing fragments of two different passages.

Additionally, `findAllElements` in `docxParser.ts` recursed into already-matched runs, so runs nested inside drawings were collected twice. The exporter (`docxExporter.ts`) and `tableSegmentation.ts` counted paragraphs with the same faulty regex, so `paragraphIndex` metadata was also misaligned for round-trip export.

## Fix

[...]

PR #1110, opened 2026-07-22
Double entries from some word files

TimRl:
## Double entries from some word files
Format:
#1109 

## Summary

Fixes duplicated content when importing `.docx` files that contain text boxes (`mc:AlternateContent`), and adds a "smart re-import" flow so projects that already imported an affected document can be repaired in place without losing translations.

**The import bug:** Word wraps text boxes in `mc:AlternateContent`, which contains the same content twice — once in `mc:Choice` (modern) and once in `mc:Fallback` (legacy). The importer read both copies, and its element scanner also recursed into nested runs (e.g. runs inside drawings inside runs), collecting the same text multiple times. Result: repeated sentences across several cells in the imported source (and therefore the target).

**The repair path:** Re-importing a file previously created a brand-new source/codex pair, so existing projects had no way to pick up the fix. Now, when an imported file matches an existing pair (by content hash, or by filename for edited documents), the user is offered "Update Existing" — the pair is rebuilt from the fresh parse while carrying translations over. The merge is fully CRDT/sync-safe: removed cells are soft-deleted tombstones (never hard-deleted), and every changed value or data field gets an explicit edit entry so changes propagate correctly through sync instead of being reverted by the merge resolver.

### Docx import (root-cause fix)
- New `ooxmlScanner.ts` utility providing canonical paragraph enumeration shared by import and export:
  - `stripFallbackElements` removes all `mc:Fallback` duplicates (handles nesting and self-closing tags).
  - `extractOutermostParagraphRanges` treats paragraphs nested inside containers (text boxes) as part of their anchor paragraph instead of counting them twice.
- `docxParser.ts`: `findAllElements` no longer recurses into already-matched elements, so nested runs are collected exactly once.
- `tableSegmentation.ts` uses the same fallback-stripped body and outermost-paragraph rules, keeping table detection consistent with paragraph enumeration.
- `docxExporter.ts` strips `mc:Fallback` before replacing content, so untranslated legacy copies can't reappear in exported files, and uses the shared paragraph enumeration so `paragraphIndex` stays consistent between import and export.

### Smart re-import ("Update Existing Import")
- Importing a file that matches an existing pair prompts: **Update Existing** or **Import as New Copy** (QuickPick if multiple pairs match). Detection uses the `originalFilesHashes` registry, falls back to scanning source metadata by hash, and matches by filename when the document was edited (different hash, same name).
- `reimportMerge.ts` (new): rebuilds the pair from the fresh parse, matching new cells to old by normalized text (exact match first, then containment for re-segmented cells). Old cell IDs and target translations are preserved; milestones are matched by chapter number so their IDs stay stable.
- Sync safety: unmatched old cells are tombstoned (`data.deleted = true` plus a `dataDeleted` edit) rather than removed; all value and data-field changes append `MIGRATION` edits; existing edit histories are preserved on both source and target. This ensures re-import survives `resolveCodexCustomMerge` when other users or the remote still have the old file.
- Dirty editors are saved before the update, and open webviews are refreshed afterwards so editors reflect the merged content immediately.
- A summary notification reports matched cells, preserved translations, and removed cells. No on-disk backup is created — a backup `.codex` copy inside the workspace gets picked up by `**/*.codex` scans (export list, migrations, indexing) and shows up as a duplicate document; recovery is covered by tombstones retaining full content and history in the file itself.
- Biblical imports whose standardized source filename already exists are excluded from re-import detection to avoid conflicting with the existing "file exists" flow.

### Export
- `docxExporter.ts` skips soft-deleted cells (`data.deleted === true`) when collecting translations, so tombstones with stale `paragraphIndex` values can't corrupt round-trip exports.

### Automated
- [x] Extension test suite passes (1304 passing).
- [x] Webview (vitest) suite passes, including new tests for `ooxmlScanner` (fallback stripping, outermost paragraph ranges, nested alternate content, self-closing tags, table cells).
- [x] New unit tests for `reimportMerge` cover: exact/containment matching, translation carry-over, tombstone semantics, `MIGRATION` edit trails, milestone ID reuse, paratext handling, and duplicate old cells favoring translated copies.
- [x] `npm run compile` clean; no linter errors on changed files.

### Manual
- [x] Imported a `.docx` with text boxes (`1 Cor L#1 Intro W.docx`) into a fresh project — previously repeated sentences (cells 6–11, 24–26) now appear exactly once.
- [x] Re-imported the same file into a project with the duplicated source — prompted "Update Existing", duplicates removed, existing translations preserved on matching cells.
- [x] Verified removed cells are tombstoned in the codex file (content and edit history intact, `data.deleted` set) rather than deleted.
- [x] Export view shows each document exactly once after an update (no duplicate entries).
- [x] Round-trip docx export after an update: tombstoned cells excluded, translated content lands in the correct paragraphs.
- [x] "Import as New Copy" still creates an independent pair when chosen.


dadukhankevin (approved):
Looks good!

PR #1115, opened 2026-08-01
fix: preserve untranslated segments of split paragraphs on DOCX expor…

### LINKED ISSUE #1113: Bug: Cells split manually with "ideal cell length" in the docx importer aren't exported correctly when empty
## Description
DOCX round-trip export loses the content of untranslated segments when a paragraph was split into multiple cells at import (e.g. via the "Ideal cell length" setting). Rejoining split segments into the original paragraph is working as designed, but segments that were never translated (their codex cells are still empty) are silently dropped during the rejoin instead of falling back to their original source text. The exported paragraph is overwritten with only the translated segments, so the untranslated portions of the original document text are erased.

Root cause, in `webviews/codex-webviews/src/NewSourceUploader/importers/docx/docxExporter.ts` (`collectTranslations()`):
- Cells with empty values are skipped (`if (!translated) continue;`), so untranslated segments never enter the per-paragraph segment list.
- When segments are rejoined (`segments.map(s => s.text).join(' ')`), there is no fallback to the segment's source text — even though every cell carries it in `metadata.data.originalText` (set in `cellMetadata.ts`).
- `replaceParagraphTextXml()` then replaces the entire original paragraph's `<w:t>` content with the partial join and empties any leftover `<w:t>` nodes, so the original text for the missing segments is destroyed rather than preserved.

The same class of bug exists in the table-cell path: a partially translated multi-paragraph table cell writes `parts[j] ?? ''` for missing lines, blanking the untranslated paragraphs.

Note the paragraph-level behavior is correct when *nothing* in the paragraph is translated (no map entry → paragraph left byte-untouched). The data loss occurs only on partial translation of a split paragraph.

## Steps to Reproduce
1. Import a .docx containing a long paragraph with the DOCX importer, setting "Ideal cell length" (Advanced settings) low enough that the paragraph splits into 3+ cells.
2. In the Codex editor, translate only one of the middle cells; leave the other segment cells empty.
3. Export using the DOCX round-trip export option.
4. Open the exported `_translated.docx` and inspect the paragraph.

## Expected Behavior
The rejoined paragraph preserves all segments in order: untranslated segments keep their original source text, and translated segments show the translation — e.g. `[source seg 0] [translation seg 1] [source seg 2]`. No original document text is lost.

## Actual Behavior
The paragraph contains only the translated segment(s) joined with spaces. The original text of the untranslated segments is gone from the exported document. In partially translated table cells, untranslated paragraphs are blanked to empty strings.

## Acceptance Criteria
- [x] When rejoining split segments in `collectTranslations()`, any segment with no translation falls back to its source text (`metadata.data.originalText`) so the full paragraph content is reconstructed.
- [x] Segment order is preserved by `segmentIndex`, including when the missing segments are at the start, middle, or end of the paragraph.
- [x] Fully untranslated split paragraphs continue to be left byte-untouched in the output (current correct behavior is not regressed).
- [x] No regression to the surgical-replacement guarantees (untouched paragraphs, styles, images, overall OOXML structure preserved; file still opens in Word and Apple Pages).

## Test Checklist
- [x] Regression: fully translated split paragraph still exports the full joined translation.
- [ ] Regression: .idml files still work.
- [x] Manual check: exported file opens correctly in Microsoft Word and Apple Pages.

Luke-Bilhorn:
## PR Title
1113-DOCX split cells aren't exported correctly when empty

## Summary
Closes #1113

DOCX round-trip export was dropping the source text of untranslated segments in a paragraph that was split at import (e.g. via "ideal cell length"). `collectTranslations()` now falls back to each cell's `data.originalText` for untranslated segments, so the full paragraph is reconstructed in `segmentIndex` order instead of being overwritten with only the translated parts.

## Changes
- Untranslated split segments fall back to `metadata.data.originalText` on rejoin (no data loss).
- Segment order preserved by `segmentIndex` (missing segments at start/middle/end).
- Fully-untranslated split paragraphs still produce no map entry → left byte-untouched.
- Table-cell path falls back to the matching original line instead of blanking untranslated paragraphs.

### Automated (docxRoundtrip.integration.test.ts)
- [x] Partial split paragraph keeps source text for untranslated segments (order preserved, only that paragraph changes)
- [x] Regression: fully-translated split paragraph exports the full joined translation
- [x] Regression: untouched paragraphs stay byte-identical

### Manual
- [x] Import .docx with low "Ideal cell length", translate one middle segment, export → untranslated segments show original text
- [x] Exported file opens correctly in Microsoft Word and Apple Pages


LeviXIII (approved):
Works ok. I'm getting the missing text. There is an issue with underlines not being under the correct characters, but that's a previous issue outside the scope of this ticket.

PR #1120, opened 2026-08-07
Fix: honor remote deletions in metadata.json merge so unpinning sticks (#1105)

### LINKED ISSUE #1105: Bug: Unpinning a project doesn't stick
## Description
Projects can be "pinned" to a specific extension version. The pin is stored in `metadata.json` under `meta.pinnedExtensions`. When an admin unpins a project, that entry gets deleted from the file, and the deletion is pushed to the server correctly. So far so good.

The problem happens on everyone else's computers. When a teammate syncs, our code has to combine three versions of `metadata.json`:

- **base** — the last version both sides agreed on
- **ours** — the teammate's local copy (still has the pin)
- **theirs** — the version from the server (pin deleted by the admin)

The merge code that combines these lives in codex-editor, in the `mergeObjects` function (`src/projectManager/utils/merge/resolvers.ts`, around line 2171). For each field it basically asks: "Did the local side change this since base? If not, use the server's value." That's the right idea, but there's a flaw in this line:

```typescript
result[key] = tVal !== undefined ? tVal : oVal;
```

`tVal` is the server's value. If the server *deleted* the field, `tVal` is `undefined` — and the code reads that as "the server has no opinion, keep the local value." It can't tell the difference between "this field was deleted on purpose" and "this field never existed." So it puts the pin back from the teammate's local copy. The teammate's client then pushes that merged file to the server, and the project is pinned again for everyone — including the admin.

Two things make this happen almost every time:

1. Whenever a teammate has *any* work that hasn't been synced yet (even one edited verse), the sync system treats every file that differs — including `metadata.json` — as a conflict that needs merging (`GitService.ts` in frontier-authentication, around line 2726). So the buggy merge runs even if the teammate never touched `metadata.json`.
2. Opening a project usually writes to `metadata.json` anyway (it records the user's app version and extension versions — see `src/utils/metadataManager.ts`), so the two sides are almost always out of sync.

One more asymmetry worth understanding: when the *local* side deleted the field, the code keeps the deletion. That's why the admin's own sync looks fine — the bug only bites when the deletion arrives *from the server*.

Note: this isn't limited to pins. Deleting *any* field from `metadata.json` (except the few fields with their own special merge logic) will quietly come back the same way. Also, these workarounds do NOT help: writing an empty `pinnedExtensions: {}` instead of deleting (the merge digs into the object and restores each entry inside), editing the file directly on GitLab (client merges undo it), or making other edits at the same time (each field merges independently). The only change that survives today is *changing* the pin's value rather than *removing* it.

## Steps to Reproduce
1. Set up a project that has `meta.pinnedExtensions` in its `metadata.json`, pushed to the server, and cloned on two machines (A = admin, B = teammate).
2. On machine B, make any small edit (one cell in any `.codex` file is enough) but don't sync yet.
3. On machine A, unpin the project (delete `meta.pinnedExtensions` from `metadata.json`) and sync. Check the server — the pin should be gone.
4. On machine B, open the project and let it sync.
5. Look at `metadata.json` on the server again, and on machine A after it next pulls.

## Expected Behavior
The pin stays deleted everywhere. Machine B's merge should recognize that the server deleted `pinnedExtensions` on purpose and keep it deleted.

## Actual Behavior
Machine B's merge silently puts `pinnedExtensions` back (copied from its local file) and pushes it to the server. The project is pinned again for everyone. There's no error or warning anywhere.

## Acceptance Criteria
- In `mergeObjects`: if the local value is unchanged from base and the server deleted the field, the merged result must leave the field out. This must work both for simple values and for whole objects (a deleted object must not be rebuilt piece by piece from its inner fields).
- If the *local* side deleted a field and the server didn't change it, the field stays deleted (this already works — don't break it).
- If *both* sides changed the same field differently, the local value still wins (existing rule — don't break it).
- The fields with their own special merge logic (`edits`, `initiateRemoteUpdatingFor`, `projectSwap`, `users`, `originalFilesHashes`) behave exactly as before.
- End to end: with the fix installed on both machines, running the steps above leaves the pin deleted on the server after machine B syncs.

## Test Checklist
- [ ] Delete all of `meta.pinnedExtensions`, local copy unchanged → merged file has no `pinnedExtensions`.
- [ ] Delete just one extension entry inside `pinnedExtensions`, local copy unchanged → only that entry is gone, the others remain.
- [ ] Local side deletes `pinnedExtensions`, server unchanged → stays deleted (protects the admin-side behavior).
- [ ] Change a pin's `version` and `url`, local copy unchanged → server's new values win (this is how pins get updated — must keep working).
- [ ] Both sides changed the same field differently → local value wins (existing rule still holds).
- [ ] Delete some other ordinary `metadata.json` field (not a pin) → deletion sticks too.
- [ ] The special-case fields (`edits`, `initiateRemoteUpdatingFor`, `projectSwap`, `users`, `originalFilesHashes`) come out the same as before the fix.
- [ ] Full run of `resolveMetadataJsonConflict` with base/ours/theirs JSON matching the unpin scenario → output has no `pinnedExtensions`.
- [ ] Manual check on two real machines following the reproduction steps, using builds that include the fix.

KaladinX:
> **Draft on purpose:** major changes are landing on `main` on Monday; this branch will ingest those before it's ready for review.

Fixes #1105.

## Problem

Unpinning a project deletes `meta.pinnedExtensions` from `metadata.json`, and the deletion reaches the server. But when a teammate with any unsynced local work then syncs, the 3-way merge in `mergeObjects` (`src/projectManager/utils/merge/resolvers.ts`) reads "field absent from theirs" as "theirs has no opinion" and keeps the local copy of the pin, which gets pushed back to the server. The project ends up pinned again for everyone, admin included, with no error anywhere. The same applies to deleting any ordinary `metadata.json` field.

## Fix

One guard in `mergeObjects`, before both the leaf merge and the object recursion: a field **present in base, absent from theirs, and unchanged locally since base** is a deliberate remote deletion, so the key is left out of the merged result. Placing it before the recursion matters — fixing only the leaf line would still rebuild a deleted object as `{}` (or entry by entry) through the recursive path, which the ticket's AC explicitly forbids.

Everything else keeps its existing behavior:

- Local changed the field, theirs deleted it → conflict, local wins (unchanged rule).
- Field added locally (absent from base *and* theirs) → kept, not mistaken for a deletion.
- Local side deleted, theirs unchanged → stays deleted (the admin-side path that already worked).
- The specialized merges (`edits`, `initiateRemoteUpdatingFor`, `projectSwap`, `users`, `originalFilesHashes`) are skipped or overridden after the generic merge, exactly as before.

`resolveMetadataJsonConflict` is now exported so tests can run the real resolver end to end, per the ticket's checklist.

## Test checklist (from #1105)

All automated items are covered by 10 new cases in `src/test/suite/resolvers/resolveMetadataJsonConflict.test.ts`:

- [x] Delete all of `meta.pinnedExtensions`, local unchanged → no `pinnedExtensions` in result
- [x] Delete one entry inside `pinnedExtensions` → only that entry gone, sibling survives
- [x] Local deletes, server unchanged → stays deleted
- [x] Server changes a pin's `version`/`url`, local unchanged → server values win
- [x] Both sides changed the same field differently → local wins
- [x] Delete an ordinary non-pin field → deletion sticks
- [x] Specialized fields still governed by their own merges (covered for `users`; the others are skipped before the guard can see them)
- [x] Full `resolveMetadataJsonConflict` run of the unpin scenario → no `pinnedExtensions`, unrelated local change still merges
- [ ] Manual check on two real machines with builds that include the fix — **still to do, needs two humans/machines**

## Riskiest part

The guard changes merge behavior for *every* field of `metadata.json` that lacks specialized handling, not just pins: remote deletions now stick when the local side hasn't touched the field. That is the ticket's intent, but it's the place to stare at in review. The comparison uses the same `JSON.stringify` equality the leaf merge already used, so "unchanged locally" is judged exactly as before.

## Gates

`npm run lint` clean; full VS Code suite: **1286 passing, 7 pending, 0 failing**.


KaladinX:
Ran an adversarial pass over this fix, tracing the pin lifecycle across frontier-authentication, the Codex shell's Conductor, and frontier-server. One thing worth a small guard before merge, one rollout caveat, and two all-clears.

**1. An empty `theirs` now wipes metadata.json (new failure mode, cheap guard available)**

> **Update:** the guard landed in c45671c4 — `theirsIsTrustworthy` (non-blank `theirs` that parses to an object with at least one key) now gates the deletion rule, each honored deletion is logged via `debugLog`, and two tests cover the empty-string and empty-object cases. The paragraphs below are kept for context.

The conflict contents come from frontier-authentication's `GitService.ts` (the conflict-building block around lines 2770-2892). `remoteContent` starts as `""`, and a failed `readBlobAtRef` on an existing file is swallowed by the surrounding try/catch. An incomplete fetch can therefore produce a conflict with `theirs: ""` and `isDeleted: false`. codex-editor's empty-content guard (the `BLOB_READ_FAILED` sentinel) only covers `isNew` files, so an existing metadata.json flows straight into the resolver.

Before this fix an empty `theirs` was harmless: every field fell through to "keep ours". With the deletion rule, `theirs` parses to `{}`, so every field the local user hasn't changed since base (projectName, languages, source/target language, and the rest) is treated as deliberately deleted, and the stripped file is committed and pushed to the whole team.

Suggested guard: disarm the deletion rule when `conflict.theirs` is blank or parses to an object with no keys. A real metadata.json always has keys, so an empty one means "the remote blob could not be read", not "the remote deleted everything". That restores the old keep-ours behavior in exactly the cases where the old behavior was the safe one. Roughly three lines in `resolveMetadataJsonConflict` plus a test.

**2. Rollout: one stale client re-pins everyone**

A teammate on an older extension still resurrects the pin and pushes it. Clients with the fix then see the pin as a remote addition (absent from base), which the merge correctly takes. So unpins only stick once every collaborator has updated, and a single stale client re-pins the project for everyone. Nothing to change in this PR; worth a line in the release notes so "the bug is back" reports during the rollout window get read correctly.

**3. Conductor cache: verified, no change needed**

I checked whether the shell's cached remote pins could keep a project pinned after metadata.json converges. They cannot: frontier-authentication clears the cache twice per sync. Pre-merge it reads pins fresh from the fetched remote head and calls `setRemotePins` with the result (`SCMManager.ts:597`), which is `undefined` after an unpin, and the Conductor removes the stored key. Post-merge it refreshes again from the merged local file (`SCMManager.ts:756`), where `readLocalPinnedExtensions` returns `undefined` when the key is absent. The planned two-machine test remains the end-to-end confirmation.

**4. Server side: all clear**

frontier-server never writes metadata.json. Its only touches are read-only GitLab raw-file fetches (`versionPins.ts` for the admin pin inventory, `metricsScheduler.ts` for validation thresholds), and no unpin endpoint exists server-side. A side benefit of this fix: the version-pins dashboard stops showing resurrected pins for projects an admin already unpinned.

PR #1121, opened 2026-08-10
feat: show timestamp overage when audio recording exceeds allowed time

Luke-Bilhorn:
## PR Title
1101-Show overage when recording audio

## Summary
Closes #1101

Shows bracketed overage (e.g. `30.0s [+5.000s]`) when recorded audio exceeds the cell's timestamp duration — in the Audio Recording tab while recording and after, and in the Timestamps tab.

## Changes
- **Audio Recording tab:** Overage in brackets next to allowed time in the Timestamp Length bar (live recorder + post-recording review)
- **Timestamps tab:** Same bracketed overage next to total audio length for the current cell range
- **Within limit:** No brackets when audio is at or under the allowed time

### Audio Recording tab
- [x] Record past allowed time → overage shows in Timestamp Length bar while recording
- [x] Same overage shows after recording stops (waveform review)
- [x] Record within allowed time → no brackets

### Timestamps tab
- [x] Overage shows in brackets next to audio length when over
- [x] No brackets when within allowed time

## Screenshots
image

image
image
image
image

Issue #1117, opened 2026-08-06
Task:  BSB stray text fixes, also their IDML files dont import extra text such as Stories of Jusus


Fikitti:
## Summary
The imported BSB files often have stray parts of bible verses imported along side of the notes, or just text that shouldnt be there at all as with NT bible.
Also, their files dont seem to import extra text thats usually at the beginning of a file, such as "Stories about Jesus" etc.

## Context
It complicated the workflow of the translators, other tools such as cell hiding are in the process, but it would be ideal if we could fine tune the import to fix those issues.

## Acceptance Criteria
- [ ] Files are imported with no stray test
- [ ] Extra content text is imported correctly and milestoned correctly

## Test Checklist
- [ ] Import all of their files from the list that have issues, and validate the functionality

PR #1122, opened 2026-08-10
ci: make macOS extension-tests non-blocking (upstream test-electron ENOENT)

Luke-Bilhorn:
## Summary
The `Node 20 • macos-latest` extension-tests check is failing on every PR (including #1121, #1105, #1117) — but it's **not a code problem**. The `@vscode/test-electron` 2.x harness downloads VS Code and then fails to launch it on GitHub's `macos-latest` (darwin-arm64) runner, **before any test executes**:

```
✔ Downloaded VS Code into .../vscode-darwin-arm64-1.132.0
Test error: Error: spawn .../Visual Studio Code.app/Contents/MacOS/Electron ENOENT
```

`lint` + `compile` pass on macOS, and the **full test suite passes on Linux and Windows**. This is an upstream harness/runner incompatibility ([microsoft/vscode-test#348](https://github.com/microsoft/vscode-test/issues/348)).

## Change
- Split the shared macOS/Windows test step so **Windows stays blocking**.
- Mark the **macOS** leg `continue-on-error: true` so it still runs (failures remain visible in the logs and the leg self-heals once fixed upstream) but no longer blocks merges.

## Why not bump `@vscode/test-electron`?
The fix lives in the 3.x line, which requires **Node >= 22**, restructures the `out/runTest` import this repo uses, and per the upstream bug may still `ENOENT` on darwin-arm64. That's a larger, coordinated toolchain change and belongs in its own PR. This unblocks all open PRs now with minimal risk.

## Revert criteria
Remove `continue-on-error` once `@vscode/test-electron` is upgraded (Node 22) and the arm64 launch is fixed upstream.

## Test plan
- [x] CI: `macos-latest` job reports green (step still runs; failure shown as "continued")
- [x] CI: `ubuntu-latest` and `windows-latest` still run and remain blocking

### DIFF
--- a/.github/workflows/extension-tests.yml
+++ b/.github/workflows/extension-tests.yml
@@ -52,8 +52,27 @@ jobs:
                   NODE_OPTIONS: --max_old_space_size=4096
               run: xvfb-run -a npm test
 
-            - name: Test (macOS/Windows)
-              if: runner.os != 'Linux'
+            - name: Test (Windows)
+              if: runner.os == 'Windows'
+              env:
+                  NODE_OPTIONS: --max_old_space_size=4096
+              run: npm test
+
+            # macOS is intentionally non-blocking for now. The VS Code test
+            # harness (@vscode/test-electron 2.x) fails to launch the downloaded
+            # darwin-arm64 build on GitHub's macos-latest runner:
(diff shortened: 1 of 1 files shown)


LeviXIII (approved):
Good fix!

PR #1127, opened 2026-08-11
Built in audio editor

LeviXIII:
## Summary

This is a PR from [linlin449](https://github.com/linlin449) for an update to audio recorder:

Adds a built-in, non-destructive audio editor to the Codex Cell Editor.

## Features

- Delete any selected audio range longer than zero seconds
- Keep audio inside a selected range
- Insert an audio file at a selected timeline position
- Drag Start, End, and Insert pointers directly on the waveform
- Precisely adjust pointer positions in 0.01-second increments
- Zoom the waveform with Ctrl + mouse wheel
- Undo delete, keep, and insert operations
- Save edited audio as a new WAV version
- Preserve the original recording
- No FFmpeg installation required

## Implementation

Audio decoding and rendering are handled inside the webview using the Web Audio API.

Edits are represented as non-destructive timeline clips. The original attachment is not modified, and a new WAV attachment is generated only when the user selects "Save as new version."

## Testing

- Extension build passed
- ESLint passed
- 1,276 extension tests passed
- Full webview test suite passed
- Added tests for:
  - Pointer range normalization
  - Equal Start and End pointers
  - Short-range deletion
  - Cross-clip deletion
  - Audio insertion
  - Browser WAV encoding
  - Undo history

### DIFF
--- a/sharedUtils/audioLimits.ts
+++ b/sharedUtils/audioLimits.ts
@@ -1,0 +1,27 @@
+/**
+ * Audio attachment size limits shared by the extension host and webviews.
+ * The provider enforces the cap when saving; webview editors use the same
+ * value to warn before rendering instead of failing after.
+ */
+
+/** Hard cap the provider enforces on a decoded audio attachment. */
+export const MAX_AUDIO_ATTACHMENT_BYTES = 50 * 1024 * 1024;
+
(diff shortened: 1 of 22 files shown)


LeviXIII (commented):
Works well!

PR #1145, opened 2026-08-21
Tools alert

TimRl:
## Summary

Added per-user native tool status tracking to `metadata.json`. The current tool status and status history are stored directly as metadata snapshots without using `editMap`.

## Changes

- Added `meta.nativeToolStatus` for the latest tool status per username.
- Added `meta.nativeToolStatusHistory` to record status changes over time.
- Records:
  - `username`
  - `timestamp`
  - `nonNative`
- Only creates a new history entry when a user’s `nonNative` tool set changes.
- Removed `editMap` handling for native tool status.
- Added conflict resolution based on username and timestamp.
- Preserves different users’ statuses during metadata merges.
- Keeps the `nonNative` array in a consistent order.
- Does not add or require `userId`.
- Avoids rewriting metadata when the status has not changed.

## Test Checklist

- [x] Verified native tool status is stored per username.
- [x] Verified unchanged `nonNative` configurations do not create duplicate history entries.
- [x] Verified status changes create new history entries with updated timestamps.
- [x] Verified native status arrays are normalized consistently.
- [x] Verified metadata conflicts merge different users independently.
- [x] Verified the newest timestamp wins for the same user.
- [x] Verified legacy native-tool `editMap` entries no longer drive status merging.
- [x] Ran lint checks on the changed files.
- [x] Ran `npm run compile` successfully.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "activationEvents": [
         "onCustomEditor:newSourceUploaderProvider-scheme",
         "onCustomEditor:startupFlowProvider-scheme",
-        "*"
+        "onStartupFinished"
     ],
     "main": "./out/extension.js",
     "permissions": [
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "activationEvents": [
         "onCustomEditor:newSourceUploaderProvider-scheme",
         "onCustomEditor:startupFlowProvider-scheme",
-        "onStartupFinished"
+        "*"
     ],
     "main": "./out/extension.js",
     "permissions": [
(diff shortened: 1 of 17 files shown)


LeviXIII (commented):
If I change to fallback tools, the "nonNative": [] doesn't get populated in either nativeToolStatus nor nativeToolStatusHistory.

TimRl:
> If I change to fallback tools, the "nonNative": [] doesn't get populated in either nativeToolStatus nor nativeToolStatusHistory.

So, the status is only taken at startup, and this accounts for us being able to turn it off temporarily, but when codex is restarted (or a project is closed and reopened) it always reverts to the native tools unless locked into fallback mode (or incompatible due to OS or architecture). So it's only during the restart if the native tools can't be loaded that it notes the fallback tools use.

If you think it'd be better to immediately reflect the status, then I can change it.

LeviXIII (approved):

LeviXIII:
Understanding it better, it is good as it is.

PR #1154, opened 2026-08-31
fix: converge identical-file syncs and safely preserve merged content

TimRl:
## Summary

Sync could repeatedly process thousands of unchanged media pointers because it treated changes in Git history as content conflicts.

This PR skips verified identical files while still completing the history merge, so subsequent syncs can finish cleanly.

## Changes

- Compare committed, working, and remote bytes before recovering missing conflicts.
- Preserve safeguards for genuinely missing changes and unsupported binary content.
- Complete divergent histories even when no files need rewriting.
- Pass the analysed commit snapshot to Frontier and retry if history changes.
- Apply three-way pointer resolution and reject competing replacements.
- Write merged content atomically to prevent readers seeing partial files.

### Companion changes in frontier-authentication

The complete fix spans both repositories. The companion `fix-streaming-sync` branch:

- Prevents unnecessary media downloads during staging.
- Stages the resolved pointer instead of retaining the old pointer.
- Prevents empty local commits.
- Validates merge snapshots and downloaded media.
- Limits concurrent conflict reads.

Both sets of changes should ship together for the complete fix.

## Test Checklist

- [x] 97 editor tests pass.
- [x] TypeScript, webpack build, and focused lint checks pass.
- [x] Regression coverage includes 1,500 identical pointers.
- [x] Genuine missing changes, unstaged edits, binary differences, and older Frontier responses remain protected.
- [x] Pointer resolution and atomic notebook writes are covered.
- [ ] Verify with an affected production project.

### DIFF
--- a/src/projectManager/utils/merge/conflictSynthesis.ts
+++ b/src/projectManager/utils/merge/conflictSynthesis.ts
@@ -1,7 +1,7 @@
 /**
- * Last-resort self-heal for the remote-changed/conflict-list invariant
- * (issue #991, enforced in merge/index.ts): every remote-changed file must
- * appear in the conflict list, or the merge commit would silently drop it.
+ * Recover remote changes missing from Frontier's conflict list (issue #991).
+ * A history change is not necessarily a content conflict: paths already
+ * present with identical committed AND working bytes need no resolution.
  *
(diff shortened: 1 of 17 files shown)


TimRl:
New branch and PR since this was a broad fix of other issues (which I figured out after having the broken project)

PR #1158, opened 2026-09-01
fix: stop synthesizing omitted sync conflicts

TimRl:
## Summary

Prevent Codex from turning diagnostic paths omitted from Frontier’s conflict list into fabricated conflicts. This previously caused remote files to be materialized, processed, and potentially committed as local work.

This PR works with the corresponding Frontier changes, which provide authoritative conflict information and the Git snapshot used during analysis.

## Changes

- Use `remoteChangedFilePaths` as the authoritative conflict-list diagnostic.
- Stop using `allChangedFilePaths`, which also contains local-only changes.
- Retry safely when Frontier’s remote-change diagnostics disagree with its conflict list.
- Never synthesize missing conflicts or write guessed remote content into the working tree.
- Complete divergent histories even when Frontier resolved every change and Codex has zero file resolutions.
- Pass the analyzed local, remote, and merge-base commits back to Frontier during merge completion.
- Optimize stream-only post-sync cleanup by filtering existing pointer stubs before replacement.
- Preserve compatibility cleanup when upload information is empty or omitted.
- Continue processing media explicitly reported as uploaded.

## Test Checklist

- [x] Path normalization covers separators, prefixes, and Unicode.
- [x] Missing conflict entries stop safely without synthesizing files.
- [x] A large omitted remote set is never materialized.
- [x] Empty and omitted LFS upload reports retain compatibility cleanup.
- [x] Existing pointer stubs are excluded from expensive replacement.
- [x] Explicitly uploaded media is still converted back to pointers.
- [x] `npm run compile`
- [x] `npm run lint`
- [x] Extension test suite: 1,513 passing, 10 pending

### DIFF
--- a/src/projectManager/utils/merge/conflictSynthesis.ts
+++ b/src/projectManager/utils/merge/conflictSynthesis.ts
@@ -1,228 +1,63 @@
-/**
- * Last-resort self-heal for the remote-changed/conflict-list invariant
- * (issue #991, enforced in merge/index.ts): every remote-changed file must
- * appear in the conflict list, or the merge commit would silently drop it.
- *
- * A mismatch is first treated as transient — the sync retries with a fresh
- * fetch. But if it persists on the final attempt (a path-classification bug
- * upstream, version skew), failing forever would cut the user off from pushing
- * AND pulling. Instead, the missing paths get ConflictFile entries built from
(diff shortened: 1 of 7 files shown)


LeviXIII (approved):
Syncing at same time works as well.

Issue #1160, opened 2026-09-03
Task: Investigate slowdown when generating Sparkle results


LeviXIII:
## Summary
Sparkle (AI translation suggestions) has become noticeably slower to generate results. Investigate what is causing the slowdown and fix it, or report back on what we find (BGP).

## Context
A partner reported a significant slowdown when generating Sparkle results while translating the NEW ACT-REV batch. On the same day, Patrice reported similar slowness while working on the front and back matter and the Bible Dictionary. Since two people saw this in different parts of the project at the same time, it is likely a general performance issue rather than something specific to one book or file.

One thing worth checking: a large batch of edits was recently copied into the NEW ACT-REV files as part of the migration work (the latest round finished in early September). We don't know yet if that is related, but the timing lines up, so it should be ruled in or out.

## Acceptance Criteria
- [ ] Reproduce the slowdown (or measure Sparkle generation times) in an affected project
- [ ] Identify the cause of the slowdown
- [ ] Confirm whether the recent NEW ACT-REV migration edits are related or not
- [ ] Fix the issue, or write up findings if the fix needs its own ticket
- [ ] Sparkle results come back at normal speed for the affected users

## Test Checklist
- [ ] Verified Sparkle generation speed while translating cells in the NEW ACT-REV files
- [ ] Verified Sparkle generation speed in the front and back matter
- [ ] Verified Sparkle generation speed in the Bible Dictionary
- [ ] Confirmed with the partner and Patrice that the slowdown is resolved

PR #1163, opened 2026-09-03
fix: generate structured IDML translations in one request

TimRl:
## Summary

Addresses an identified cause of slow Sparkle generation in multi-segment IDML cells: plain-text translations could not preserve the imported structure, triggering an additional LLM request to repair the HTML.

Eligible cells now request indexed JSON translations and reconstruct the original HTML locally, avoiding the repair request on successful structured responses.

Related to #1160. Broader project-wide performance verification remains pending.

## Changes

- Extract source text segments and their `data-segment-index` values.
- Request schema-constrained JSON translations through `response_format`.
- Validate segment counts, indexes, duplicates, and translation types before reconstruction.
- Reinsert escaped translations while preserving original tags, attributes, styles, and separators.
- Retain the legacy fallback for unsupported providers or invalid responses; cancellation does not trigger a retry.
- Limit the new path to supported multi-segment IDML cells with HTML structure enforcement enabled. Other formats, single-segment cells, and merged/range cells retain existing behavior.
- Add regression coverage for reconstruction, validation, fallback, cancellation, and compatibility.

### Automated verification

- [x] Extension build passes.
- [x] ESLint passes.
- [x] Full extension test suite passes: 1,530 passing, 10 pending.
- [x] Valid structured output completes with one LLM request.
- [x] Reconstruction preserves markup and correctly maps out-of-order segment responses.
- [x] Invalid responses are rejected and the legacy fallback remains functional.
- [x] Cancellation does not start a fallback request.
- [x] Single-segment cells and disabled structure enforcement retain the legacy path.

### Performance and acceptance verification

- [x] Reviewed the supplied IDML recording: translation appears after approximately 13–14 seconds, without the additional HTML-repair stage.
- [x] Verify generation speed in NEW ACT–REV.
- [x] Verify representative front/back matter and Bible Dictionary cells.
- [x] Confirm whether recent migration edits contributed to the reported slowdown.
- [ ] Confirm normal generation speed with the partner and Patrice.

### DIFF
--- a/src/providers/codexCellEditorProvider/codexCellEditorProvider.ts
+++ b/src/providers/codexCellEditorProvider/codexCellEditorProvider.ts
@@ -4401,7 +4401,11 @@ export class CodexCellEditorProvider implements vscode.CustomEditorProvider<Code
                         completionConfig,
                         cancellationToken,
                         true, // returnHTML
-                        isBatchOperation
+                        isBatchOperation,
+                        {
+                            enforceHtmlStructure:
+                                currentDocument.getNotebookMetadata().enforceHtmlStructure === true,
+                        },
                     );
 
                     // Check for cancellation before updating document - this is crucial to prevent cell population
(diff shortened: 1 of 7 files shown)


LeviXIII (approved):
Speed looks normal to me. Tried on a couple of different projects.

PR #1170, opened 2026-09-10
Edit history freely floating

### LINKED ISSUE #1169: Bug: Edit History panel stays fixed in the middle of the screen instead of staying with its cell
## Description
When you open the Edit History for a cell (the clock icon in the cell editor), the panel appears in the middle of the editor window. If you then scroll the list of cells, the panel does not move with the cell it belongs to. It stays pinned to the same spot on the screen and covers whatever cells scroll underneath it.

In the screenshot this was reported with, the cell being edited (cell 2) has scrolled partly out of view, while the Edit History panel is still sitting over cells 3 and 4, which have nothing to do with it. It is easy to lose track of which cell the history belongs to.

The panel should be attached to the cell whose history was opened, so it scrolls together with that cell.

Note for developers: the panel is currently positioned relative to the window (fixed, centred at 50% / 50%) rather than relative to the cell. See `webviews/codex-webviews/src/CodexCellEditor/Editor.tsx` around line 1765.

Seen in a The Chosen project (episode 102, Mongolian Cyrillic) in the `.codex` translation editor with the source text open side by side, but this is not specific to that project.

## Steps to Reproduce
1. Open any project and open a `.codex` file for translation.
2. Click a cell to start editing it.
3. Click the clock icon (Show Edit History) in the cell's toolbar. The Edit History panel opens.
4. Scroll the list of cells up or down with the mouse wheel or trackpad.

## Expected Behavior
The Edit History panel stays attached to the cell it was opened from. When you scroll, the panel moves up and down together with that cell, and it never covers unrelated cells.

## Actual Behavior
The panel stays fixed in the middle of the screen while the cells scroll behind it. The cell it belongs to can scroll away entirely while the panel stays put, covering other cells.

## Acceptance Criteria
- [ ] The Edit History panel opens directly next to (or below) the cell whose history was requested, and it is clear which cell it belongs to.
- [ ] Scrolling the cell list moves the panel together with its cell. The panel never floats over other cells.
- [ ] If the cell scrolls out of view, the panel goes with it. No panel is left behind in the middle of the screen.
- [ ] Long edit histories can still be read and scrolled through inside the panel.
- [ ] The panel still closes with the X button and with the Escape key.
- [ ] The "History" and "LLM Previews" tabs both stay attached to the cell.
- [ ] The panel works the same whether the source text is open side by side or the translation editor fills the window.

### Panel follows the cell
- [ ] Open a cell for editing, open its Edit History, scroll up and down. The panel stays with the cell.
- [ ] Scroll far enough that the cell leaves the visible area. The panel is no longer visible either, and nothing is left floating on screen.
- [ ] Scroll back. The panel is still open and still attached to the same cell.

### Different cell positions
- [ ] Open the Edit History on the first cell in a chapter. The panel is fully readable and not cut off.
- [ ] Open the Edit History on the last cell in a chapter. The panel is fully readable and not cut off.
- [ ] Open the Edit History on a cell with many edits (a long history). The list scrolls inside the panel and the panel does not become too tall to use.

### Closing and tabs
- [ ] Close the panel with the X button.
- [ ] Close the panel with the Escape key.
- [ ] Switch between the "History" and "LLM Previews" tabs, then scroll. The panel stays attached on both tabs.

### Layouts
- [ ] Test with the source text open side by side (split view) and with the translation editor filling the whole window.
- [ ] Test in a narrow window to make sure the panel does not overflow past the edge of the editor.

Image

TimRl:
## Summary
Closes #1169

Attach Edit History to its cell editor instead of the viewport. The panel overlays the editor without increasing its height, shrinks to fit short histories, and scrolls internally for longer histories.

## Changes
- Position the panel within its cell so it moves with the cell and leaves the viewport with it.
- Constrain the overlay to the editor’s content area.
- Keep both History and LLM Previews in the same attached panel.
- Preserve existing X button and Escape key handling.
- Use the editor’s width so the panel adapts to split and full-width layouts.
- Fix DOCX hashing compatibility with older Node versions, resolving the unrelated test failures blocking the push.

### Automated verification
- [x] Cell editor production build passes.
- [x] Extension build and lint pass.
- [x] Extension tests pass: 1,687 passing.
- [x] Webview tests pass: 837 passing, 3 skipped.
- [x] Push succeeds with pre-push checks enabled.

### Manual verification still required
- [x] Scroll the cell out of view and back; history stays attached.
- [x] Verify first and last cells have fully readable panels.
- [x] Confirm short histories fit their content and long histories scroll internally.
- [x] Close the panel using X and Escape.
- [x] Switch between History and LLM Previews and verify scrolling.
- [x] Verify split view, full-width view, and narrow windows.


Luke-Bilhorn (approved):
Looks good works good. Only thing I couldn't test was imported target edits displaying no history.

PR #1172, opened 2026-09-12
1168 character metadata audio export

### LINKED ISSUE #1168: Feature: Store the character on each cell so "Consolidate by Character" groups by character, not by label text
## Summary
Give each cell its own **character** field, separate from the display label. The Cell Label Importer fills it in from the character column of the script spreadsheet. The **Consolidate by Character** audio export then groups lines by that field instead of by the full label text.

Today the export treats every different label as a different character. In a project where labels include the camera position (for example `simon_on`, `simon_off`, `simon_mixed`), Simon's lines are split across three files instead of one.

## Problem
**Current workflow.** A dubbing team imports a script spreadsheet with the Cell Label Importer. They pick one or more columns to build the label shown on each cell. Later they run Export Project with **Consolidate by Character** to get one audio file per character for their audio editing software (DAW).

**Why it does not work well.** The export has no idea what a character is. It only knows the label. If the label carries anything besides the character name, such as a camera position, the export makes one file per label variation. All the audio is exported, but the team has to merge the files by hand.

**What already exists.** The importer already reads a character column from the spreadsheet (a `character` or `CHARACTER` column, or `CHARACTER LABEL 1`, `2`, and so on) and shows it in its preview table. It then discards that value and saves only the combined label. So the information we need is already there at import time. It is just not kept.

**Why not strip keywords from the label.** We could teach the export to remove words like "on", "off", or "mixed". That breaks the moment a team writes `SIMON (CAMERA)` or any other wording. A separate character field avoids guessing at the shape of the label entirely. The label can say whatever the team wants.

**Follow-up (not part of this issue).** Projects that were imported before this change, or that have hand-typed labels, will not have the field. A later issue can add a step in the "Preview characters" dialog where the user ticks several labels and names them as one character, saved in project metadata. For now, re-running the importer with the character column selected is the way to fill in the field on an existing project.

## Acceptance Criteria 
- [ ] Each cell can hold a character value that is separate from its display label.
- [ ] In the Cell Label Importer, the user can pick which spreadsheet column is the character, independently of the columns used to build the label.
- [ ] Running the importer with a character column selected fills in the character on every matched cell without changing the label.
- [ ] Re-running the importer on an existing project with a character column selected fills in the character on existing cells and leaves their labels as they are.
- [ ] The character value syncs between team members like any other cell edit.
- [ ] **Consolidate by Character** produces one audio file per character value. Lines with labels `simon_on`, `simon_off`, and `simon_mixed` and character `SIMON` all land in one file named with `simon`.
- [ ] Cells with no character value fall back to being grouped by their label, so projects that never set a character export exactly as they do today.
- [ ] The **Preview characters** list matches the files that will be exported: one entry per character.

### Cell Label Importer
- [ ] Import a spreadsheet with a `CHARACTER` column and a `CAMERA` column. Build the label from both columns and choose `CHARACTER` as the character. Confirm the label on each cell shows both values and the character shows only the name.
- [ ] Import a spreadsheet with no character column selected. Confirm the import still works and no character is set.
- [ ] Re-run the importer on a project that already has labels, selecting only the character column. Confirm the labels do not change and the character is now filled in.
- [ ] Make a change on one machine and sync. Confirm the character value appears on a second machine after it syncs.

### Consolidate by Character export
- [ ] On a project with `simon_on`, `simon_off`, and `simon_mixed` labels and character `SIMON`, run the export and confirm only one Simon file is produced.
- [ ] Open that file in an audio editor and confirm lines from all three camera positions are present at the right times.
- [ ] Click **Preview characters** and confirm Simon appears once.
- [ ] On a project where no cell has a character value, run the export and confirm the files match what the export produces today.
- [ ] On a project where only some cells have a character value, confirm those cells group by character and the rest group by label.
- [ ] Confirm with the reporter that the exported files now drop into their DAW as expected.

TimRl:
# #1168 — Character audio export options

## Summary
Closes #1168

Combine camera-angle variants into one audio track per character during export. By default, labels such as `Alex (ON)` and `Alex (Off)` consolidate as `Alex`, preserving the recordings’ timeline positions. Users can customize the ignored markers or keep the full labels as separate tracks.

## Changes
- Ignore trailing `(ON)`, `(Off)`, `(Mixed)`, and `(Group)` markers by default, without modifying notebook labels or importing separate character metadata.
- Add a **Character options…** modal to keep the export card compact.
- Add **Separate by Camera Angles per Character** to preserve full labels. Hide the ignored-marker controls when enabled and explain that no markers are ignored.
- Allow users to add or remove literal trailing markers, including their parentheses. An empty list preserves all markers.
- Add an optional **Case sensitive** checkbox, unchecked by default.
- Remember custom markers and case sensitivity locally for the current project/workspace. Reset defaults restores the four markers and case-insensitive matching.
- Apply the same grouping settings to preview, audio-only export, and audio included with other export formats.
- Preserve existing audio formats, timeline alignment, and trimming behavior.
- Fix DOCX buffer handling encountered during validation: pass byte-array views to Web Crypto and ZIP loading, with regression coverage for cross-context buffers and SHA-256 correctness.

### Character grouping and preview
- [x] Default markers combine into the same character group regardless of letter case.
- [x] Case-sensitive matching removes only markers with matching capitalization.
- [x] Camera-angle separation preserves full labels.
- [x] Custom markers, an empty list, repeated suffixes, and unrecognized parentheses behave as expected.
- [x] Preview applies the selected grouping and case-sensitivity options.
- [x] Preview leaves stored notebook labels unchanged.
- [x] All 8 character grouping and preview tests pass after the case-sensitivity change.

### Options interface
- [x] Generated-UI checks verify defaults, checkbox changes, settings messages, and reset behavior.
- [x] Marker controls and reset are hidden when camera-angle separation is enabled.
- [x] Toggling camera-angle separation preserves custom markers.
- [x] Reset defaults clears case-sensitive matching.
- [x] Confirm the final modal layout and saved preferences after installing the final VSIX and restarting VS Code.

### Audio output
- [x] Sample WAV, FLAC, and Opus exports combine all four default camera-angle variants into one track.
- [x] Decoded sample audio retains recordings at their original timestamps, silent gaps, and expected duration.

### Build and regression checks
- [x] Extension compile and lint pass after adding case-sensitive matching.
- [x] DOCX regression coverage verifies cross-context input produces the expected SHA-256 and readable document content.
- [x] Latest full webview run: 838 passed, 3 skipped; uploader production build passed.
- [x] Earlier full extension run for the export-grouping implementation: 1,694 passed, 10 pending.

The full-suite results above precede the final case-sensitivity addition; that addition was checked with compile, lint, generated-UI checks, and the 8 focused grouping tests. The final checkbox-label edit was checked with `git diff --check`.

## 1168 character metadata audio export

Format:
#1168

## Summary
Closes #1168 

Describe the change.

## Changes

## Test Checklist
If generating from Claude, please do the checklist yourself before submitting.
You may use sub-headers to organize better as follows:

### Header 1
- [ ] It does the thing

### Header 2
- [ ] It doesn't do the other thing

## Screenshots
Only if necessary for clarity


Luke-Bilhorn (approved):
Kieran says good!

